PR #50, opened 2025-07-17
Add `kagenti-client-registration`

Alan-Cha:
Add kagenti-client-registration, which enables Kagenti to automatically register clients (agents and tools) with Keycloak


cwiklik (changes requested):
The initContainer image must be in external repo otherwise agent/tool deployment will fail after application installation. Current code depends on locally built image which is loaded into Kind cluster.

Alan-Cha:
@cwiklik I now point at https://github.com/orgs/kagenti/packages/container/package/kagenti-client-registration so you shouldn't need to build the image anymore.

PR #53, opened 2025-07-28
Make KEYCLOAK_URL optional for kagenti-client-registration

Alan-Cha:
* Make KEYCLOAK_URL optional for client registration (simply does not attempt registration with Keycloak)


platform-operator/internal/deployer/kubernetes/kubernetes.go
```diff
@@ -298,13 +303,10 @@ func (d *KubernetesDeployer) createDeployment(ctx context.Context, component *pl
...
-		 	{
-				Name:  "CLIENT_NAME",
-				Value: clientName,
-			}
-		}
-	)
+	mainEnvs = append(mainEnvs, corev1.EnvVar{
+		Name:  "CLIENT_NAME",
+		Value: component.Name,
+	})
```
Alan-Cha, line 309: There seemed to be an error here? `append()` expects an element, not an array.

Alan-Cha: This might have been fixed.

cwiklik (approved):
LGTM

PR #58, opened 2025-08-26
:book: Using correct apiVersion on Agent kind

matzew:
## Summary

Fixing little typo on `apiVersion` for the `Agent` kind

## Related issue(s)

Fixes #


cwiklik:
The beeai operator has reached End of Life and is no longer supported.

PR #66, opened 2025-09-02
🌱 add automation to tag latest release

pdettori:
## Summary

Should help with automating retrieving latest operator release without using GH API.


pdettori:
seems like we can just use a different approach based on a command such as `git ls-remote --tags  --sort=-version:refname https://github.com/kagenti/kagenti-operator.git`

PR #77, opened 2025-09-24
Run spiffe-helper to allow client registration to use SVID

Alan-Cha:


platform-operator/internal/deployer/kubernetes/kubernetes.go
```diff
@@ -246,74 +246,6 @@ func (d *KubernetesDeployer) createDeployment(ctx context.Context, component *pl
...
-					Name:  "CLIENT_ID",
-					Value: "spiffe://localtest.me/sa/" + component.Name,
-				},
-				{
-					Name:  "NAMESPACE",
-					Value: namespace,
-				},
-			},
-		})
-	}
```
Alan-Cha, line 316: There are no more initContainers

This is because client registration now depends on spiffe-helper and spiffe-helper is a long running process so it cannot be an initContainer.

platform-operator/internal/deployer/kubernetes/kubernetes.go
```diff
@@ -347,8 +279,100 @@ func (d *KubernetesDeployer) createDeployment(ctx context.Context, component *pl
...
+										Name: "environments",
+									},
+									Key: "KEYCLOAK_ADMIN_PASSWORD",
+								},
+							},
+						},
+						{
+							Name:  "CLIENT_NAME",
+							Value: component.Name,
+						},
```
Alan-Cha, line 366: Note that `CLIENT_ID` and `NAMESPACE` env vars have been removed

platform-operator/internal/deployer/kubernetes/kubernetes.go
```diff
@@ -347,8 +279,100 @@ func (d *KubernetesDeployer) createDeployment(ctx context.Context, component *pl
...
+							Name: "KEYCLOAK_ADMIN_PASSWORD",
+							ValueFrom: &corev1.EnvVarSource{
+								ConfigMapKeyRef: &corev1.ConfigMapKeySelector{
+									LocalObjectReference: corev1.LocalObjectReference{
+										Name: "environments",
+									},
+									Key: "KEYCLOAK_ADMIN_PASSWORD",
+								},
+							},
+						},
```
Alan-Cha, line 362: We still need the credentials

PR #107, opened 2025-10-17
Enable Slack demo to work with SPIFFE ID

Alan-Cha:


platform-operator/internal/deployer/kubernetes/kubernetes.go
```diff
@@ -264,10 +264,6 @@ func (d *KubernetesDeployer) createDeployment(ctx context.Context, component *pl
 			Name:  "CLIENT_NAME",
 			Value: clientId,
 		},
-		{
-			Name:  "CLIENT_ID",
-			Value: "spiffe://localtest.me/sa/" + component.Name,
-		},
```
Alan-Cha: `CLIENT_ID` should not be provided as an env var. Instead, it's extracted from the mounted SVID.

platform-operator/internal/deployer/kubernetes/kubernetes.go
```diff
@@ -302,6 +312,7 @@ func (d *KubernetesDeployer) createDeployment(ctx context.Context, component *pl
 					VolumeMounts: append(
 						component.Spec.Deployer.Kubernetes.VolumeMounts,
 						sharedMount,
+						SVIDMount,
```
Alan-Cha: Mount the SVID volume so that the agent/tool can get the `CLIENT_ID`.

PR #137, opened 2025-11-24
🐛 Move CRDs to a templates subdirectory

usize:
Helm will not add subchard crds in a /crds directory in an upgrade.

This is a known limitation: https://github.com/helm/helm/issues/9538


pdettori:
closing since it is outdated but feel free to reopen if updated

PR #160, opened 2026-01-28
✨  AgentCard → Workload Identity Binding via CRD Extension

kevincogan:
## Summary

Implements **AgentCard → Workload Identity Binding via CRD Extension (Step 1)**: policy-based binding between an `AgentCard` and a workload identity derived from Kubernetes metadata.

**RFC:** [AgentCard → Workload Identity Binding via CRD Extension (Step 1)](https://docs.google.com/document/d/1sOpE9tcw-DlG4Gi31t8e_CV15lNwpZ-NyA2OCDeZZEU/edit?usp=sharing)

## Related issue(s)

Fixes [#429 - feature: A Registry Binding A2A Agent Cards and Workload Identity](https://github.com/kagenti/kagenti/issues/429)

## Motivation

Without binding, any workload can claim any AgentCard. This change adds a Kubernetes-native mechanism to declare and enforce which workload identity is allowed to serve a card.

## Example

```yaml
spec:
  identityBinding:
    allowedSpiffeIDs:
      - "spiffe://cluster.local/ns/prod/sa/weather-sa"
    strict: true # Scale to 0 on mismatch
```

## Design

- **Identity model:** a workload identity represented as a SPIFFE ID derived from Kubernetes metadata  
  `spiffe://<trust-domain>/ns/<namespace>/sa/<serviceAccount>`

- **Policy location:** the allowlist lives on the AgentCard `AgentCard.spec.identityBinding.allowedSpiffeIDs`

- **Evaluation:** the AgentCard controller computes `expectedSpiffeID` from the **Agent resource** it targets (Agent namespace + the ServiceAccount on the Agent’s pod template). This does **not** require pods to be running.

- **Enforcement (strict mode):** if `strict=true` and the binding check fails, the Agent controller **disables the Agent** by scaling its Deployment to `0`. When the binding is fixed (all strict bindings pass), it restores the Deployment back to the previous replica count.

- **Observability:** results are written to status fields and emitted as Kubernetes Events; ambiguous selectors (where an AgentCard matches multiple Agents) are detected and reported.

## Key Changes

- **API / CRDs**
  - **AgentCard:** adds `identityBinding` policy + publishes binding results in status
  - **Agent:** adds enforcement status to record when/why an Agent was disabled and what replicas to restore

- **Controller behavior**
  - **AgentCard controller:** evaluates binding and reports status + Events
  - **Agent controller:** enforces strict mode (scale to 0) and restores automatically + Events

- **Docs**
  - Quickstart guide + sample manifests

## Events

| Event | Meaning |
|------|---------|
| `BindingEvaluated` | Binding passed |
| `BindingFailed` | Binding failed |
| `BindingEnforced` | Strict mode enforced (deployment scaled to 0) |
| `BindingRestored` | Restored after binding passes |
| `MultipleAgentsMatched` | Selector matched more than one Agent |

## How to Test

- **Unit Test**: `go test ./internal/controller/... -run Identity -v`
- **Integration Test**: `go test ./test/integration/... -run IdentityBinding -v`
- **Manual Test**: follow `docs/identity-binding-quickstart.md` (match, mismatch, strict enforce, restore, ambiguous selector)

## Notes

- **Backwards compatible**: `identityBinding` is optional.
- **Step 2 (future)**: runtime cryptographic enforcement via mTLS/mesh or gateway.

### DIFF
--- a/kagenti-operator/api/v1alpha1/agentcard_types.go
+++ b/kagenti-operator/api/v1alpha1/agentcard_types.go
@@ -56,6 +56,15 @@ type IdentityBinding struct {
 	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9]([a-zA-Z0-9\-\.]*[a-zA-Z0-9])?$`
 	TrustDomain string `json:"trustDomain,omitempty"`
 
+	// ExpectedSpiffeID overrides the auto-derived SPIFFE ID for this workload.
+	// Use this when your SPIRE configuration uses a non-standard identity format
+	// (e.g., pod labels, container names, or other workload attestor selectors).
+	// If not specified, the controller derives the ID as:
+	//   spiffe://<trust-domain>/ns/<namespace>/sa/<serviceAccount>
+	// See: https://github.com/spiffe/spire/blob/main/doc/plugin_agent_workloadattestor_k8s.md
+	// +optional
+	ExpectedSpiffeID SpiffeID `json:"expectedSpiffeID,omitempty"`
+
 	// AllowedSpiffeIDs is the allowlist of SPIFFE IDs that can bind to this agent.
 	// Each ID must be a valid SPIFFE ID in the format spiffe://<trust-domain>/<path>
 	// +required
--- a/kagenti-operator/config/crd/bases/agent.kagenti.dev_agentcards.yaml
+++ b/kagenti-operator/config/crd/bases/agent.kagenti.dev_agentcards.yaml
@@ -88,6 +88,16 @@ spec:
                       type: string
                     minItems: 1
                     type: array
+                  expectedSpiffeID:
+                    description: |-
+                      ExpectedSpiffeID overrides the auto-derived SPIFFE ID for this workload.
+                      Use this when your SPIRE configuration uses a non-standard identity format
+                      (e.g., pod labels, container names, or other workload attestor selectors).
+                      If not specified, the controller derives the ID as:
+                        spiffe://<trust-domain>/ns/<namespace>/sa/<serviceAccount>
+                      See: https://github.com/spiffe/spire/blob/main/doc/plugin_agent_workloadattestor_k8s.md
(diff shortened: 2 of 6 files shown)


kagenti-operator/api/v1alpha1/agentcard_types.go
```diff
@@ -30,6 +30,31 @@ type AgentCardSpec struct {
...
+	// +optional
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9]([a-zA-Z0-9\-\.]*[a-zA-Z0-9])?$`
+	TrustDomain string `json:"trustDomain,omitempty"`
+
+	// AllowedSpiffeIDs is the allowlist of SPIFFE IDs that can bind to this agent.
+	// Each ID must be a valid SPIFFE ID in the format spiffe://<trust-domain>/<path>
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	AllowedSpiffeIDs []string `json:"allowedSpiffeIDs"`
+
```
usize: Even though we don't have the sigstore infra yet, I think we need to have

`ExpectedSignature *AgentCardSignature `json:"expectedSignature,omitempty"``

Where AgentCardSignature is from the spec [here](https://a2a-protocol.org/latest/specification/#447-agentcardsignature).

Otherwise our agents can still serve any card they want, and we'll just note that the content changed.

This way, we can supply a signature here--even if it's manually added for now--and provide a guarantee that the spiffe id matches it.

e.g.

```
  if expectedSignature != nil {
      bound = verifySignature(cardContent, expectedSignature) &&
              spiffeId in allowedSpiffeIDs
  }
```

cardId remains useful for observability, but the binding decision should require signature verification when configured.

The way you're handling enforcement and SPIFFE Ids is good, we just seem to be missing the binding piece. Right?

usize (approved):

pdettori:
@kevincogan  pls. see [instructions](https://github.com/kagenti/kagenti-operator/pull/160/checks?check_run_id=62271244487) from failing CI DCO check to fix it

Issue #161, opened 2026-01-31
epic: Refactor AgentCard controller to work with Deployments and StatefulSets


pdettori:
As part of the epic https://github.com/kagenti/kagenti/issues/523 we need to refactor AgentCard controller to work with Deployments and StatefulSets.

This is also tracked in https://github.com/kagenti/kagenti/issues/555

# Description

 ## Overview

  Migrate the AgentCard controller from watching the custom Agent CRD to watching standard Kubernetes workloads (Deployments and StatefulSets). This aligns with the Kagenti UI migration
  where agents are now deployed directly as Deployments/StatefulSets rather than through the Agent CRD.

##  Motivation

  - Remove dependency on Agent CRD for agent card indexing
  - Support agents deployed as standard Kubernetes workloads
  - Enable extensibility to any workload type (Job, CronJob, etc.) without code changes
  - Follow Kubernetes conventions (Gateway API pattern)

##  Approach: targetRef with Duck Typing

  Instead of using label selectors, the AgentCard will use an explicit targetRef field:

```
  apiVersion: agent.kagenti.dev/v1alpha1
  kind: AgentCard
  metadata:
    name: my-agent-card
  spec:
    targetRef:
      apiVersion: apps/v1
      kind: Deployment       # or StatefulSet, Job, etc.
      name: my-agent
```

##  Benefits:
  - Explicit reference to exactly one workload (no ambiguity)
  - Extensible to any workload type without code changes
  - Uses unstructured.Unstructured for duck typing
  - Follows Gateway API's PolicyTargetReference pattern

##  Design Document

  See [kagenti/docs/plans/migrate-agent-crd-to-workloads.md](https://github.com/kagenti/kagenti/blob/main/docs/plans/migrate-agent-crd-to-workloads.md) for the full migration plan.

  ---
##  Phases
  
  - [x] #163
  - [x] #164
    - Add targetRef field with apiVersion, kind, and name
    - Deprecate label-based selector field (keep for backward compatibility)
    - Update validation to require either targetRef or selector
    - Add print columns for workload type and name
    - Update CRD documentation
  - [x] #165
    - Add getWorkloadByTargetRef() method using unstructured.Unstructured
    - Implement readiness detection for Deployments and StatefulSets
    - Update protocol label lookup to support both kagenti.io/protocol and legacy kagenti.io/agent-protocol
    - Fall back to selector if targetRef is not specified
    - Add watches for Deployments and StatefulSets
    - Update RBAC permissions
[...]

Issue #116, opened 2025-10-22
feature: Strict CardSignature Checking


usize:
### Feature Description

Introduce a --require-a2a-signature flag to enforce verification of A2A AgentCard signatures before allowing agent-to-agent communication.

When enabled:

- The gateway and/or controller verify incoming AgentCards against a set of trusted public keys stored in Kubernetes Secrets (PEM or JWKS).

- If a signature is missing or cannot be verified, the peer connection is blocked.

- Enforcement can be implemented via ingress/egress NetworkPolicies or mesh AuthorizationPolicies.

Goal:

To ensure that only agents with authenticated, signed AgentCards can be discovered or connected to, preventing spoofed or unverified peers.

### Proposed Solution

 - New --require-a2a-signature CLI flag available on gateway/controller.

 - Trusted key material loaded from Kubernetes Secret(s).

 - Verification performed per A2A spec https://a2a-protocol.org/latest/specification/#556-agentcardsignature-object 

 - Unverified peers blocked via network or gateway policy.

 - Audit mode and metrics for verification results.

### Want to contribute?

- [x] I would like to work on this issue.

### Additional Context

Related to: kagenti/kagenti#303

kevincogan:
Hi @usize, I’ve implemented something similar in the past and would love to work on this issue. Can I be assigned to this issue/feature?

PR #176, opened 2026-02-11
✨ A2A AgentCard Signature Verification

kevincogan:
## Summary

Adds cryptographic signature verification for A2A AgentCards. This enables operators to ensure that AgentCard data has not been tampered.

### New Features

- **Signature verification** with two provider backends:
  - **Secret provider**: public keys stored in Kubernetes Secrets (with key-rotation fallback)
  - **JWKS provider**: public keys fetched from a remote JWKS endpoint

- **Canonical JSON hashing**: using a struct-to-map approach that strips the `signature` field and empty values before signing, future-proof against field additions

- **Algorithm support**: RSA (PKCS1v15 / SHA-256) and ECDSA (P-256, P-384, P-521) for both Secret and JWKS providers

- **Audit mode**: for gradual rollout, logs verification failures as warnings but allows cards to sync

- **Prometheus metrics**: (`signature_verification_total`, `signature_verification_duration_seconds`, `signature_verification_errors_total`) for monitoring verification outcomes

- **NetworkPolicy enforcement** : `AgentCardNetworkPolicyReconciler` creates/removes NetworkPolicies based on signature verification status

- **SignatureIdentityMatch**: combined check that both signature verification and identity binding pass, stored in `AgentCardStatus`

- **Helm values**: for configuring signature verification (`signatureVerification.enabled`, `providerType`, `secretName`, `auditMode`, etc.)

- **Kubernetes Events**: emitted for `SignatureEvaluated` (Normal) and `SignatureFailed` (Warning), visible via `kubectl describe agentcard`

### Architecture Alignment

- Signature verification works with `targetRef` (duck-typed workloads: Deployments, StatefulSets, Legacy Agent CRDs)
- NetworkPolicy controller adapted to the Deployment/StatefulSet architecture, resolving workloads via `spec.targetRef` → `status.targetRef` → `selector` fallback
- New status fields added to `AgentCardStatus`: `validSignature`, `signatureVerificationDetails`, `signatureKeyId`, `signatureIdentityMatch`
- Updated RBAC for `secrets` (get/list/watch) and `networking.k8s.io/networkpolicies`
- Deep copy generated for `CardSignature` struct

## Related issue(s)

[feature: Strict CardSignature Checking #161 ](https://github.com/kagenti/kagenti-operator/issues/116)

### Unit & Integration Tests

```bash
cd kagenti-operator

# Signature package unit tests (38 tests)
go test ./internal/signature/... -v -count=1

# Controller integration tests signature scenarios (6 tests, requires envtest)
go test ./internal/controller/... -v -count=1 --ginkgo.focus="Signature"
```

### End-to-End Demo

Follow the step-by-step guide in [`docs/a2a-signature-verification.md`](kagenti-operator/docs/a2a-signature-verification.md) which covers:

1. **Setup**: kind cluster, cert-manager, operator build & deploy
2. **Demo 1**: Valid signed AgentCard (verified)
3. **Demo 2**: Unsigned card rejected (blocked)
4. **Demo 3**: Wrong-key signature rejected (blocked)
5. **Demo 4**: Audit mode (warns but allows)
6. **Demo 5**: JWKS provider verification
7. **Demo 6**: Zero-downtime key rotation


kagenti-operator/docs/a2a-signature-verification.md
```diff
@@ -0,0 +1,811 @@
...
+```python
+#!/usr/bin/env python3
+"""Sign an A2A AgentCard with RSA private key."""
+import json, base64, sys
+from datetime import datetime, timezone
+from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives.asymmetric import padding
+from cryptography.hazmat.backends import default_backend
+
+def sign_agent_card(card_path, private_key_path, key_id):
```
usize: Right now we're able to verify that a card's content hasn't changed from an expected value, AND we're able to detect that our SPIFFE ID is one that we expect. We still need to put them together.

We just need to add one more tiny thing here: a JWS protected header with the SPIFFE ID in it. Then the signature itself will contain the svid.

Looking into it, that aligns with the A2A spec's signature format (see [here](https://a2a-protocol.org/latest/specification/#842-signature-format)).

So we'd go from:

  sig = private_key.sign(canonical.encode('utf-8'), ...)
  card['signature'] = {'algorithm': 'RS256', 'keyId': key_id, 'value': base64(sig)}

  to:

  protected = {"alg": "RS256", "typ": "JOSE", "kid": key_id, "spiffe_id": spiffe_id}
  protected_b64 = base64url(json.dumps(protected, separators=(',', ':')))
  payload_b64 = base64url(canonical)
  sig = private_key.sign((protected_b64 + '.' + payload_b64).encode('ascii'), ...)
  card['signatures'] = [{"protected": protected_b64, "signature": base64url(sig)}]

For this diff, the SPIFFE ID would be manually provided as an argument. But ideally, in a follow-up diff, we can come up with an init container that a CI system can use to add it AND sign it using SPIRE instead of providing our own JWKS and secret.

Then we'd have a very tidy binding that doesn't require input from the agent author.

kagenti-operator/api/v1alpha1/agentcard_types.go
```diff
@@ -199,6 +212,30 @@ type AgentCardData struct {
 	// SupportsAuthenticatedExtendedCard indicates if the agent has an extended card
 	// +optional
 	SupportsAuthenticatedExtendedCard *bool `json:"supportsAuthenticatedExtendedCard,omitempty"`
+
+	// Signature contains the A2A CardSignature for verification
+	// +optional
+	Signature *CardSignature `json:"signature,omitempty"`
```
usize: Calling out again here, that I goof'd. We need this split up like in the spec so that we can leverage the protected header: https://a2a-protocol.org/latest/specification/#842-signature-format

PR #269, opened 2026-04-10
Add configurable SPIFFE ID template support

akram:
## Summary

Fixes #268

The client registration controller was hardcoding the SPIFFE ID format, but ClusterSPIFFEID resources allow custom `spiffeIDTemplate` configurations. This caused confusion when operator-generated Keycloak client IDs didn't match the actual SPIFFE IDs issued to workloads.

## Changes

- Add `--spire-id-template` flag with sensible default: `spiffe://{{.TrustDomain}}/ns/{{.Namespace}}/sa/{{.ServiceAccount}}`
- Support `SPIRE_ID_TEMPLATE` environment variable for configuration (flag takes precedence)
- Refactor `resolveKeycloakClientID()` to use `text/template` for rendering SPIFFE IDs
- Add test coverage for custom template formats
- Update documentation in `operator-managed-client-registration.md`

## Template Variables

The template supports:
- `{{.TrustDomain}}` - SPIRE trust domain
- `{{.Namespace}}` - Kubernetes namespace
- `{{.ServiceAccount}}` - Service account name

## Configuration Precedence

1. Command-line flag `--spire-id-template` (highest)
2. Environment variable `SPIRE_ID_TEMPLATE`
3. Default template (if neither is set)

## Test plan

- [x] Unit tests pass (`TestResolveKeycloakClientID`)
- [x] Custom template rendering verified
- [x] Build succeeds after rebase onto latest main
- [x] Manual testing: Deploy with custom ClusterSPIFFEID template and verify client ID matches

### DIFF
--- a/kagenti-operator/cmd/main.go
+++ b/kagenti-operator/cmd/main.go
@@ -89,6 +89,7 @@ func main() {
 	var enableMLflow bool
 
 	var spireTrustDomain string
+	var spireIDTemplate string
 	var spireTrustBundleConfigMapName string
 	var spireTrustBundleConfigMapNS string
 	var spireTrustBundleConfigMapKey string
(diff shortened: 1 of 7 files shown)


grs:
This won't work as the ID need not be restricted to these three data items. If a different format is set, ot most likely will include some other element(s) (e.g. pod name, image etc).

akram:
per discussion, it seems indeed that this would be insufficient way of solving the issue. closing

PR #272, opened 2026-04-11
fix: prevent cyclic pod restarts from resign-trigger

pdettori:
## Summary

- Fix cyclic pod restarts caused by the `resign-trigger` annotation (every ~30 min)
- Add `resign-leaf-expiry` annotation to track which cert expiry already triggered a restart
- Skip SVID expiry restart if the current cert's `LeafNotAfter` matches the stored value

## Root Cause

The SVID expiry check (`time.Until(leafNotAfter) < grace`) and the cooldown guard (`time.Since(lastTrigger) < grace`) both used the same 30-minute grace period. With SPIRE's default 1-hour SVID TTL, this created a guaranteed restart cycle: the cooldown expired at exactly the moment the fresh cert entered the grace window.

```
t=0:      Pod starts, SVID valid until t+60min
t=30min:  Cooldown expires. SVID has ~29.5min left < 30min grace → RESTART
t=30min:  New pod starts, fresh SVID valid until t+90min
t=60min:  Cooldown expires. New SVID has ~29.5min left < 30min grace → RESTART
...forever
```

## Fix

Store the `LeafNotAfter` of the cert that prompted the restart in a new pod-template annotation (`agentcard.kagenti.dev/resign-leaf-expiry`). On subsequent reconciles, skip the SVID expiry trigger if the current cert has the same expiry. Once the pod restarts with a fresh SVID (different `LeafNotAfter`), the guard passes and the new cert is checked normally.

The trust-bundle rotation path is unaffected.

## Test plan

- [x] New test: "should not trigger restart when the same cert expiry was already handled" — reproduces the cyclic restart
- [x] New test: "should trigger restart when cert is renewed with a new expiry that is also near grace" — verifies genuine re-triggers work
- [x] New test: "should store resign-leaf-expiry when triggering SVID restart" — verifies annotation storage
- [x] All 151 existing controller tests pass (zero regressions)
- [ ] CI pipeline green

Fixes #246

### DIFF
--- a/kagenti-operator/internal/controller/agentcard_controller.go
+++ b/kagenti-operator/internal/controller/agentcard_controller.go
@@ -75,6 +75,11 @@ const (
 	// when the operator detects that the signing SVID is expiring or the CA has rotated.
 	AnnotationResignTrigger = "agentcard.kagenti.dev/resign-trigger"
 
+	// AnnotationResignLeafExpiry records the LeafNotAfter of the cert that prompted the
+	// last resign-triggered restart. Prevents cyclic restarts when the SVID TTL is ≤ 2×grace:
+	// the operator skips re-triggering if the current cert's expiry matches the stored value.
+	AnnotationResignLeafExpiry = "agentcard.kagenti.dev/resign-leaf-expiry"
+
 	// AnnotationBundleHash records the trust bundle hash at the time of the last signing.
 	AnnotationBundleHash = "agentcard.kagenti.dev/bundle-hash"
 
@@ -1153,6 +1158,12 @@ func (r *AgentCardReconciler) maybeRestartForResign(ctx context.Context, agentCa
 	reason := ""
 
 	if !vr.LeafNotAfter.IsZero() && time.Until(vr.LeafNotAfter) < grace {
+		// Skip if we already triggered a restart for this exact cert expiry.
+		// This prevents cyclic restarts when SVID TTL ≤ 2×grace (issue #246).
+		storedExpiry := podAnnotations[AnnotationResignLeafExpiry]
+		if storedExpiry == vr.LeafNotAfter.Format(time.RFC3339) {
+			return
+		}
 		needsRestart = true
 		reason = fmt.Sprintf("SVID leaf cert expiring at %s", vr.LeafNotAfter.Format(time.RFC3339))
 	}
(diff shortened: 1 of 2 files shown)


kagenti-operator/internal/controller/agentcard_controller.go
```diff
@@ -1153,6 +1158,12 @@ func (r *AgentCardReconciler) maybeRestartForResign(ctx context.Context, agentCa
 	reason := ""
 
 	if !vr.LeafNotAfter.IsZero() && time.Until(vr.LeafNotAfter) < grace {
+		// Skip if we already triggered a restart for this exact cert expiry.
+		// This prevents cyclic restarts when SVID TTL ≤ 2×grace (issue #246).
+		storedExpiry := podAnnotations[AnnotationResignLeafExpiry]
+		if storedExpiry == vr.LeafNotAfter.Format(time.RFC3339) {
+			return
+		}
 		needsRestart = true
```
cwiklik, line 1167: **praise**: Clean early-return guard comparing stored vs current `LeafNotAfter`. This doesn't touch the bundle rotation path at all — the `resign-leaf-expiry` annotation is only written when `!leafNotAfter.IsZero()`, so bundle-hash-triggered restarts are unaffected.

kagenti-operator/internal/controller/proactive_restart_test.go
```diff
@@ -80,6 +80,32 @@ func getResignTrigger(ctx context.Context, name, ns string) string {
...
+	if d.Spec.Template.Annotations == nil {
+		return ""
+	}
+	return d.Spec.Template.Annotations[AnnotationResignLeafExpiry]
+}
+
+func setResignTriggerAndLeafExpiry(ctx context.Context, name, ns, triggerTime, leafExpiry string) {
+	Eventually(func() error {
+		d := &appsv1.Deployment{}
+		if err := k8sClient.Get(ctx, types.NamespacedName{Name: name, Namespace: ns}, d); err != nil {
```
cwiklik, line 97: **praise**: All three new test scenarios cover the exact edge cases well:
1. Same cert expiry already handled → no restart (the bug repro)
2. Different cert near grace → restart (genuine re-trigger still works)
3. Annotation storage verification

Good coverage of the fix without over-testing.

cwiklik (approved):
## Review Summary

Targeted fix for cyclic pod restarts caused by the resign-trigger mechanism (issue #246). The root cause analysis is excellent: when SVID TTL ≤ 2×grace, the cooldown and the grace window overlap perfectly, creating an infinite restart loop. The fix adds a `resign-leaf-expiry` annotation to track which cert expiry was already handled, with a clean early-return guard. Three new tests cover all scenarios. Controller change is minimal and well-scoped.

**E2E failure is unrelated** — `inject.kagenti.io` webhook returns `connection refused` when creating a curl pod in the operator namespace (pre-existing flake in the metrics test).

**Areas reviewed**: Go (controller logic, unit tests)
**Commits**: 1 commit, signed-off ✓, `Assisted-By` convention ✓
**CI status**: Unit Tests pass ✓, E2E fails (pre-existing flake, unrelated)

**Verdict**: APPROVE

PR #282, opened 2026-04-14
fix(mlflow): fall back to Service URL when status.address.url is not set

Ygnas:
## Summary
The RHOAI MLflow operator does not always populate status.address.url on the MLflow CR. When the CR is Available but the field is empty, discover the MLflow Service by name and construct the in-cluster URL from its namespace.

Store the resolved URI in ResolveTrackingURI so CRD/Service lookups only happen once, eliminating per-reconcile log noise.

Fixes #

### DIFF
--- a/kagenti-operator/internal/controller/mlflow_controller.go
+++ b/kagenti-operator/internal/controller/mlflow_controller.go
@@ -163,11 +163,17 @@ func (r *MLflowReconciler) mlflowClient(baseURL string) *mlflow.Client {
 }
 
 // trackingURI returns the MLflow tracking URI, using the override if set.
+// After the first successful resolution, the result is stored in
+// ResolveTrackingURI so subsequent reconciles skip CRD/Service lookups.
 func (r *MLflowReconciler) trackingURI(ctx context.Context) string {
 	if r.ResolveTrackingURI != nil {
 		return r.ResolveTrackingURI(ctx)
 	}
-	return r.resolveTrackingURI(ctx)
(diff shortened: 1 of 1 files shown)


kagenti-operator/internal/controller/mlflow_controller.go
```diff
@@ -184,12 +190,29 @@ func (r *MLflowReconciler) resolveTrackingURI(ctx context.Context) string {
...
-				continue
+			var uri string
+			if cr.Status.Address != nil && cr.Status.Address.URL != "" {
+				uri = cr.Status.Address.URL
+			} else {
+				// Fallback: the MLflow CRD is cluster-scoped so GetNamespace() is
+				// empty. Discover the namespace by finding the Service that matches
+				// the CR name.
+				svcList := &corev1.ServiceList{}
+				if err := r.List(ctx, svcList); err == nil {
```
ChristianZaccaria, line 201: It's best to avoid listing all services cluster-wide, and instead list by namespace or field selector or label selector to narrow down the results.

ChristianZaccaria, line 201: We should log the error so they're visible in operator logs.
```suggestion
				if err := r.List(ctx, svcList); err != nil {
				    logger.Error(err, "failed to list Services for MLflow tracking URI fallback", "cr", cr.GetName())
				} else {


```

ChristianZaccaria (commented):
Nice catch Ignas, this is great! Left a couple of nits.

PR #296, opened 2026-04-18
fix: collision-free port assignment for proxy-sidecar mode

huang195:
## Summary

In proxy-sidecar mode, inbound traffic must go through the authbridge reverse proxy for JWT validation. Without iptables interception, the Service targetPort determines whether traffic hits the reverse proxy or goes directly to the agent.

The webhook now uses **port-stealing**: the reverse proxy takes over the agent's original port (e.g., `:8000`), and the agent is moved to a free port. This way the Service targetPort stays correct without any Service patching.

## Problem

When deploying an agent with `kagenti.io/authbridge-mode: proxy-sidecar`, the hardcoded port assignments (8080 reverse proxy, 8081 forward proxy) could collide with ports already in use by the agent, causing bind failures at runtime.

## Fix

The webhook now performs collision-free port assignment:

1. Collects all ports in use across all pod containers
2. Steals the agent's first port for the reverse proxy (so the Service doesn't need patching)
3. Finds the first free port for the agent (starting from original+1)
4. Finds the first free port for the forward proxy (starting from 8081)
5. Injects REVERSE_PROXY_ADDR, REVERSE_PROXY_BACKEND, FORWARD_PROXY_ADDR env vars into the authbridge-proxy container
6. Injects HTTP_PROXY/HTTPS_PROXY with the actual forward proxy port into app containers
7. Patches the agent's PORT env var to bind to the new port

If no free port is available (port exhaustion), the webhook rejects the pod with a clear admission error.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/webhook/injector/...` passes
- [ ] E2E: deploy agent with proxy-sidecar mode, verify inbound auth logs show `inbound authorized`

### DIFF
--- a/kagenti-operator/internal/webhook/injector/pod_mutator.go
+++ b/kagenti-operator/internal/webhook/injector/pod_mutator.go
@@ -270,17 +270,95 @@ func (m *PodMutator) InjectAuthBridge(ctx context.Context, podSpec *corev1.PodSp
 	if authBridgeMode == ModeProxySidecar {
 		// Proxy-sidecar mode: inject lightweight authbridge container + HTTP_PROXY env vars.
 		// No iptables, no proxy-init, no Envoy.
+		//
+		// Port-stealing: the reverse proxy takes over the agent's original port so
+		// the Service doesn't need patching. The agent is moved to a free port.
+		//   Service → :8000 → reverse proxy (validates JWT) → :8002 → agent
+		//   Agent outbound → HTTP_PROXY=127.0.0.1:8081 → forward proxy
+
+		// Collect all ports in use across all containers in the pod.
+		usedPorts := map[int32]bool{}
+		for _, c := range podSpec.Containers {
+			for _, p := range c.Ports {
+				usedPorts[p.ContainerPort] = true
+			}
+		}
+
+		// Find the first app container's port and steal it for the reverse proxy.
+		var originalAgentPort int32
+		var agentContainer *corev1.Container
+		for i := range podSpec.Containers {
+			c := &podSpec.Containers[i]
+			if c.Name == AuthBridgeProxyContainerName ||
+				c.Name == SpiffeHelperContainerName ||
+				c.Name == ClientRegistrationContainerName {
+				continue
+			}
+			if len(c.Ports) > 0 {
+				originalAgentPort = c.Ports[0].ContainerPort
+				agentContainer = c
+				break
+			}
+		}
+
+		if originalAgentPort == 0 {
+			originalAgentPort = 8000
+			mutatorLog.Info("no container port found, using default", "port", originalAgentPort)
+		}
+
+		// Find free ports for the agent and forward proxy.
+		// findFreePort returns the first port >= start that isn't in usedPorts,
+		// and marks it as used. Respects the valid port range (1-65535).
+		findFreePort := func(start int32) int32 {
+			p := start
+			for usedPorts[p] && p < 65535 {
+				p++
+			}
+			usedPorts[p] = true
+			return p
+		}
+
+		// Reserve the original agent port for the reverse proxy
+		usedPorts[originalAgentPort] = true
+
+		newAgentPort := findFreePort(originalAgentPort + 1)
+		forwardProxyPort := findFreePort(8081)
+
+		// Move the agent to the free port.
+		// Most agent frameworks (Python/uvicorn, Node/express, FastAPI) read the
+		// PORT env var to determine the bind address. Go agents that hardcode their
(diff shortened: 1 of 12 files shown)


kagenti-operator/test/e2e/fixtures.go
```diff
@@ -808,7 +808,7 @@ spec:
 // --- AuthBridge Injection E2E fixtures ---
```
pdettori, line 808: nit: Comment says "3 ConfigMaps" but the list now has 4 (`authbridge-config`, `authbridge-runtime-config`, `spiffe-helper-config`, `envoy-config`).

kagenti-operator/internal/webhook/injector/pod_mutator.go
```diff
@@ -270,17 +270,106 @@ func (m *PodMutator) InjectAuthBridge(ctx context.Context, podSpec *corev1.PodSp
...
+			for _, p := range c.Ports {
+				usedPorts[p.ContainerPort] = true
+			}
+		}
+
+		// Find the first app container's port and steal it for the reverse proxy.
+		var originalAgentPort int32
+		var agentContainer *corev1.Container
+		for i := range podSpec.Containers {
+			c := &podSpec.Containers[i]
```
pdettori, line 291: suggestion: Consider adding a test for the port exhaustion path (where `findFreePort` returns an error). The error handling is correct, but having test coverage would validate the admission rejection behavior.

kagenti-operator/internal/webhook/injector/pod_mutator.go
```diff
@@ -270,17 +270,106 @@ func (m *PodMutator) InjectAuthBridge(ctx context.Context, podSpec *corev1.PodSp
...
+		// the Service doesn't need patching. The agent is moved to a free port.
+		//   Service → :8000 → reverse proxy (validates JWT) → :8002 → agent
+		//   Agent outbound → HTTP_PROXY=127.0.0.1:8081 → forward proxy
+
+		// Collect all ports in use across all containers in the pod.
+		usedPorts := map[int32]bool{}
+		for _, c := range podSpec.Containers {
+			for _, p := range c.Ports {
+				usedPorts[p.ContainerPort] = true
+			}
```
pdettori, line 284: suggestion: Consider a test for the fallback path: pod with no container ports (uses default 8000, `agentContainer == nil`). The warning log is good, but a test assertion would lock down the behavior.

kagenti-operator/internal/webhook/injector/pod_mutator.go
```diff
@@ -270,17 +270,106 @@ func (m *PodMutator) InjectAuthBridge(ctx context.Context, podSpec *corev1.PodSp
...
+				c.Name == ClientRegistrationContainerName {
+				continue
+			}
+			if len(c.Ports) > 0 {
+				originalAgentPort = c.Ports[0].ContainerPort
+				agentContainer = c
+				break
+			}
+		}
+
```
pdettori, line 303: praise: Clean port reservation pattern — `findFreePort` marks allocated ports as used, preventing subsequent calls from reusing them. Handles the tricky case where agent port and forward proxy port could collide.

kagenti-operator/internal/webhook/injector/pod_mutator.go
```diff
@@ -270,17 +270,106 @@ func (m *PodMutator) InjectAuthBridge(ctx context.Context, podSpec *corev1.PodSp
...
+		// Move the agent to the free port.
+		// Most agent frameworks (Python/uvicorn, Node/express, FastAPI) read the
+		// PORT env var to determine the bind address. Go agents that hardcode their
+		// listen port won't be affected by this env var — they must use PORT or
+		// be configured via their own config mechanism.
+		if agentContainer != nil {
+			agentContainer.Ports[0].ContainerPort = newAgentPort
+			setOrAddEnv(agentContainer, "PORT", fmt.Sprintf("%d", newAgentPort))
+			mutatorLog.Info("proxy-sidecar port stealing",
```
pdettori, line 346: praise: Good fix — now skips HTTP_PROXY injection for spiffe-helper and client-registration sidecars, not just authbridge-proxy.

pdettori (approved):
Clean port-stealing implementation with collision detection. The core `findFreePort` closure correctly scans and reserves ports, preventing cascading collisions. The ConfigMap rename from `authbridge-unified-config` to `authbridge-runtime-config` is consistent across all 12 files. All 15 CI checks passing including E2E.

**Areas reviewed**: Go, Helm, Tests, Docs, E2E fixtures
**Commits**: 5 commits, all signed-off with DCO
**CI status**: All 15 checks passing

PR #297, opened 2026-04-20
feat: Generate per-agent ConfigMap for authbridge at injection time

huang195:
## Summary

- Operator creates a per-agent ConfigMap (`authbridge-config-<name>`) at pod injection time
- Reads the namespace-level `authbridge-runtime-config`, merges in mode + per-agent listener addresses
- Authbridge sidecar mounts the per-agent ConfigMap instead of the shared one
- Removes `REVERSE_PROXY_ADDR`, `REVERSE_PROXY_BACKEND`, `FORWARD_PROXY_ADDR` env vars and `--mode` flag from the proxy-sidecar container (all now in the per-agent ConfigMap)

## Context

Listener addresses are per-agent (port-stealing assigns dynamic ports) but the shared ConfigMap is per-namespace. Previously we used `${ENV_VAR}` placeholders in the shared ConfigMap, which broke validation when env vars were unset in other modes.

The three-layer ConfigMap flow:
1. **Helm** creates `authbridge-runtime-config` per namespace (defaults)
2. **Backend** creates `authbridge-runtime-config` for runtime namespaces (same defaults)
3. **Operator** reads it at injection, merges per-agent config, writes `authbridge-config-<agent>`

## Files changed

| File | Change |
|------|--------|
| `namespace_config.go` | Read `authbridge-runtime-config` ConfigMap |
| `resolved_config.go` | Propagate `AuthBridgeRuntimeYAML` |
| `pod_mutator.go` | `ensurePerAgentConfigMap()` + call sites for both modes |
| `volume_builder.go` | Parameterize authbridge ConfigMap name, `overrideAuthBridgeConfigMapInVolumes` |
| `container_builder.go` | Remove listener env vars and `--mode` from proxy-sidecar |
| `*_test.go` | Updated tests for new behavior |

## Test plan

- [x] `go test ./internal/webhook/injector/` — all pass
- [x] `go vet` — clean
- [x] `gofmt` — clean
- [ ] E2E: deploy agent with default mode → per-agent ConfigMap with `mode: envoy-sidecar`
- [ ] E2E: deploy with `kagenti.io/authbridge-mode=proxy-sidecar` → per-agent ConfigMap with listener addresses

### DIFF
--- a/kagenti-operator/internal/webhook/injector/pod_mutator.go
+++ b/kagenti-operator/internal/webhook/injector/pod_mutator.go
@@ -667,9 +667,15 @@ func (m *PodMutator) ensurePerAgentConfigMap(
 		}
 		// Does not exist — create
 		if createErr := m.Client.Create(ctx, cm); createErr != nil {
-			return "", fmt.Errorf("failed to create ConfigMap %s/%s: %w", namespace, cmName, createErr)
+			if apierrors.IsAlreadyExists(createErr) {
+				// Race: another pod's admission created it between our Get and Create.
+				mutatorLog.Info("Per-agent ConfigMap created by concurrent admission", "namespace", namespace, "name", cmName)
+			} else {
+				return "", fmt.Errorf("failed to create ConfigMap %s/%s: %w", namespace, cmName, createErr)
+			}
+		} else {
+			mutatorLog.Info("Created per-agent ConfigMap", "namespace", namespace, "name", cmName, "mode", mode)
 		}
-		mutatorLog.Info("Created per-agent ConfigMap", "namespace", namespace, "name", cmName, "mode", mode)
 	} else {
 		// Exists — update if managed by us
 		if existing.Labels[managedByLabel] != managedByValue {
--- a/kagenti-operator/internal/webhook/injector/pod_mutator.go
+++ b/kagenti-operator/internal/webhook/injector/pod_mutator.go
@@ -102,6 +102,8 @@ func NewPodMutator(
 	}
 }
 
+// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=create;get;list;update;watch
+
 // InjectAuthBridge evaluates the multi-layer precedence chain and conditionally injects sidecars.
 //
 //nolint:gocyclo // sequential injection steps form a single logical pipeline
--- a/kagenti-operator/internal/webhook/injector/pod_mutator.go
+++ b/kagenti-operator/internal/webhook/injector/pod_mutator.go
@@ -407,7 +407,9 @@ func (m *PodMutator) InjectAuthBridge(ctx context.Context, podSpec *corev1.PodSp
 			podSpec.Containers = append(podSpec.Containers, builder.BuildClientRegistrationContainerWithSpireOption(crName, namespace, spireEnabled))
 		}
 
-		// Inject volumes — use per-agent ConfigMap name for authbridge config
+		// Inject volumes — use per-agent ConfigMap name for authbridge config.
+		// requiredVolumes is always set above (resolved or legacy path) before
+		// the mode switch, so it is never nil here.
 		proxyVolumes := overrideAuthBridgeConfigMapInVolumes(requiredVolumes, perAgentCMName)
 		for i := range proxyVolumes {
 			if !volumeExists(podSpec.Volumes, proxyVolumes[i].Name) {
@@ -692,7 +694,9 @@ func (m *PodMutator) buildOwnerReference(ctx context.Context, namespace, crName 
 			WithAPIVersion("apps/v1").
 			WithKind("Deployment").
 			WithName(deploy.Name).
-			WithUID(deploy.UID)
+			WithUID(deploy.UID).
+			WithController(true).
+			WithBlockOwnerDeletion(true)
 	}
 
 	// Try StatefulSet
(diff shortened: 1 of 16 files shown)


esnible:
I asked Claude about this PR, and it believes there are problems:

```
⏺ Found a critical issue — the RBAC is missing the patch verb for configmaps, but server-side apply uses Patch. Let me verify what the kubebuilder
  marker generates and what the Helm chart has.                                                                                                                                                                                                         
                                                                                          
⏺ The RBAC is missing patch in both the kubebuilder marker, the kustomize role.yaml, and the Helm role.yaml. Server-side apply (client.Patch)      
  requires the patch verb. This is a must-fix bug.  
```

kagenti-operator/internal/webhook/injector/pod_mutator.go
```diff
@@ -508,6 +549,167 @@ func (m *PodMutator) ensureServiceAccount(ctx context.Context, namespace, name s
...
+	}
+
+	// Try StatefulSet
+	sts := &appsv1.StatefulSet{}
+	if err := m.Client.Get(ctx, client.ObjectKey{Namespace: namespace, Name: crName}, sts); err == nil {
+		return applyconfigsmetav1.OwnerReference().
+			WithAPIVersion("apps/v1").
+			WithKind("StatefulSet").
+			WithName(sts.Name).
+			WithUID(sts.UID)
```
pdettori: **Suggestion**: `buildOwnerReference` doesn't set `Controller: true` or `BlockOwnerDeletion: true`. GC works without these, but K8s convention is to mark the primary managing controller with `controller=true`. Without it, a second controller could also set a controller OwnerReference on this CM (K8s only rejects a second `controller=true` reference, not a second `controller=nil/false` one).

Suggested change for both Deployment and StatefulSet paths:
```go
WithController(ptr.To(true)).
WithBlockOwnerDeletion(ptr.To(true))
```

kagenti-operator/internal/webhook/injector/pod_mutator.go
```diff
@@ -380,16 +407,18 @@ func (m *PodMutator) InjectAuthBridge(ctx context.Context, podSpec *corev1.PodSp
 			podSpec.Containers = append(podSpec.Containers, builder.BuildClientRegistrationContainerWithSpireOption(crName, namespace, spireEnabled))
 		}
 
-		// Inject volumes (shared-data, authbridge-runtime-config, spire volumes if enabled)
-		for i := range requiredVolumes {
-			if !volumeExists(podSpec.Volumes, requiredVolumes[i].Name) {
-				podSpec.Volumes = append(podSpec.Volumes, requiredVolumes[i])
+		// Inject volumes — use per-agent ConfigMap name for authbridge config
+		proxyVolumes := overrideAuthBridgeConfigMapInVolumes(requiredVolumes, perAgentCMName)
```
pdettori, line 409: **Suggestion**: `requiredVolumes` is only set inside the `PerWorkloadConfigResolution` block (a few lines up). If `PerWorkloadConfigResolution=false` and proxy-sidecar mode is active, `requiredVolumes` is nil here — `overrideAuthBridgeConfigMapInVolumes` returns an empty slice and the per-agent CM name never makes it into the injected volumes.

If proxy-sidecar always requires `PerWorkloadConfigResolution=true` (i.e. the scenario isn't reachable in practice), a short comment here prevents future confusion:
```go
// requiredVolumes is set in the PerWorkloadConfigResolution branch above;
// proxy-sidecar mode always requires that gate.
```
Otherwise, the volume override should also cover the non-resolved path.

kagenti-operator/test/e2e/e2e_test.go
```diff
@@ -577,6 +573,36 @@ var _ = Describe("AuthBridge Injection E2E", Ordered, func() {
...
 
+			By("collecting diagnostic logs before envoy admin test")
+			diagCmd := exec.Command("kubectl", "logs", podName,
+				"-n", authBridgeTestNamespace,
+				"-c", "envoy-proxy", "--tail=30")
+			if envoyLogs, diagErr := utils.Run(diagCmd); diagErr == nil {
+				_, _ = fmt.Fprintf(GinkgoWriter, "=== envoy-proxy logs (last 30) ===\n%s\n", envoyLogs)
+			}
+			diagCmd = exec.Command("kubectl", "get", "configmap",
+				"authbridge-config-authbridge-agent",
```
pdettori: **Nit**: CM name constructed by hand. If the test agent name changes this silently diverges. Since `perAgentConfigMapName` is unexported, consider a helper constant:
```go
const testAgentCMName = "authbridge-config-" + authBridgeAgentName
```

pdettori (approved):
## Summary

Solid PR. The design is clean: lifting per-agent concerns out of the shared ConfigMap into a dedicated `authbridge-config-<name>` CM at injection time is the right call, and the progression from Get→Create/Update → AlreadyExists handling → server-side apply shows good iterative thinking. The uncached APIReader fix (commit `6b305bf`) is the kind of subtle cache-scope bug that usually only surfaces in production — good catch.

**Areas reviewed**: Go (injector), Helm RBAC, kustomize RBAC, unit tests, E2E tests
**Commits**: 18 commits, all signed-off ✓
**CI status**: All checks pass (Build, E2E, Unit, Integration, Lint, CodeQL, Trivy) ✓

Approve with two suggestions and one nit below.

PR #308, opened 2026-04-27
fix: Auto-create authbridge ConfigMaps in agent namespaces from templates

### LINKED ISSUE #1336: bug: authbridge-runtime-config ConfigMap missing for namespaces not in Helm values
## Problem

When agent namespaces are created outside of the Helm chart (e.g. via GitOps, the UI, or manual `kubectl`), the `authbridge-runtime-config` ConfigMap is not created in those namespaces. The upgraded authbridge sidecar (v0.5.0-alpha.5) requires an `identity` block with `jwt_svid_path` in its config. Without `authbridge-runtime-config`, the envoy-proxy container crashes with:

```
config validation: identity.type=spiffe requires socket_path or jwt_svid_path
```

## Root cause

- The `authbridge-runtime-config` ConfigMap is created by `charts/kagenti/templates/agent-namespaces.yaml` (lines 320-352), but only for namespaces listed in `Values.agentNamespaces`
- The backend has a fallback path (`_build_authbridge_runtime_yaml` in `kagenti/backend/app/routers/agents.py`) that creates this ConfigMap when deploying agents via the UI
- However, the operator/webhook that reads this ConfigMap as the base config has no fallback -- it expects the ConfigMap to exist and crashes without it
- The per-agent `authbridge-config-<agent>` ConfigMaps generated by the webhook inherit from the namespace-level `authbridge-runtime-config`; when it's missing, the identity block is absent

## Steps to reproduce

1. Deploy Kagenti with SPIFFE/SPIRE enabled (`authBridge.clientAuthType: federated-jwt`)
2. Create an agent namespace that is NOT in `Values.agentNamespaces`
3. Deploy an agent in that namespace
4. Observe that the authbridge sidecar crashes with the identity validation error

## Suggested fix

1. The operator/webhook should have a defensive check: if `authbridge-runtime-config` is missing in the namespace, either create a default one or fail with a clear error message pointing the user to add the namespace to `agentNamespaces`
2. The installer should add a post-install validation step that checks all namespaces with the `kagenti-enabled` label have the required ConfigMaps
3. Consider making the webhook/operator create `authbridge-runtime-config` from cluster-level defaults if the namespace-level ConfigMap is absent

## Workaround

Manually create the ConfigMap in affected namespaces:
```bash
oc apply -f - <<CMEOF
apiVersion: v1
kind: ConfigMap
metadata:
  name: authbridge-runtime-config
  namespace: <agent-namespace>
data:
  config.yaml: |
    mode: envoy-sidecar
    inbound:
      issuer: "https://<keycloak-url>/realms/kagenti"
    outbound:
      keycloak_url: "http://keycloak-service.keycloak.svc:8080"
      keycloak_realm: "kagenti"
      default_policy: "passthrough"
    identity:
      type: "spiffe"
      client_id_file: "/shared/client-id.txt"
      client_secret_file: "/shared/client-secret.txt"
      jwt_svid_path: "/opt/jwt_svid.token"
CMEOF
```

Or add the namespace to `agentNamespaces` in values and re-run the Helm upgrade.

## Environment

- Platform: OpenShift 4.20.14
- Sidecar image: v0.5.0-alpha.5
- Auth mode: federated-jwt (SPIFFE)

huang195:
## Summary

- AgentRuntime controller now copies 4 template ConfigMaps (`authbridge-config`, `authbridge-runtime-config`, `envoy-config`, `spiffe-helper-config`) from `kagenti-system` to agent namespaces during reconciliation
- Create-if-not-exists semantics — existing ConfigMaps (user/Helm/backend-created) are never overwritten
- Templates are read via uncached APIReader (same pattern as `ClientRegistrationReconciler` and `PodMutator`)
- Non-fatal: if template creation fails, reconciliation continues and the webhook/kubelet surfaces the error at pod time

## Merge order

> **Merge the Helm chart PR first:** https://github.com/kagenti/kagenti/pull/1350
>
> That PR creates the template ConfigMaps in `kagenti-system`. This operator PR copies them to agent namespaces. The operator gracefully no-ops when templates are absent (logs at V(1) and skips), so merging in either order is safe — but the fix only works once both are in.

Fixes https://github.com/kagenti/kagenti/issues/1336 (part 2 of 2)

## Changes

| File | Change |
|------|--------|
| `internal/controller/agentruntime_controller.go` | Add `APIReader` field, `ensureNamespaceConfigMaps()`, RBAC annotation update |
| `cmd/main.go` | Wire `APIReader: mgr.GetAPIReader()` |
| `internal/controller/agentruntime_controller_test.go` | 4 test cases (create from templates, skip existing, skip when no templates, partial creation) |

## Test plan

- [x] All 155 controller tests pass
- [x] All webhook/injector tests pass
- [x] `go build ./...` succeeds
- [ ] Manual E2E: create namespace without ConfigMaps, deploy AgentRuntime, verify ConfigMaps appear

### DIFF
--- a/kagenti-operator/internal/controller/agentruntime_controller.go
+++ b/kagenti-operator/internal/controller/agentruntime_controller.go
@@ -57,8 +57,9 @@ const (
 // AgentRuntimeReconciler reconciles AgentRuntime objects.
 type AgentRuntimeReconciler struct {
 	client.Client
-	Scheme   *runtime.Scheme
-	Recorder record.EventRecorder
+	Scheme    *runtime.Scheme
+	Recorder  record.EventRecorder
+	APIReader client.Reader // uncached reader for cross-namespace ConfigMap reads
 }
 
 // +kubebuilder:rbac:groups=agent.kagenti.dev,resources=agentruntimes,verbs=get;list;watch;create;update;patch;delete
@@ -66,7 +67,7 @@ type AgentRuntimeReconciler struct {
 // +kubebuilder:rbac:groups=agent.kagenti.dev,resources=agentruntimes/finalizers,verbs=update
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;update;patch
 // +kubebuilder:rbac:groups=apps,resources=statefulsets,verbs=get;list;watch;update;patch
-// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch
+// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch
 // +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
 // +kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
 
@@ -111,6 +112,16 @@ func (r *AgentRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	r.setCondition(rt, ConditionTypeTargetResolved, metav1.ConditionTrue, "TargetFound",
 		fmt.Sprintf("%s %s resolved", rt.Spec.TargetRef.Kind, rt.Spec.TargetRef.Name))
+
+	// 4.5. Ensure required authbridge ConfigMaps exist in the namespace.
+	// Copies templates from kagenti-system if missing, matching the backend's
+	// _ensure_authbridge_configmaps() semantics (create-if-not-exists).
+	if err := r.ensureNamespaceConfigMaps(ctx, rt.Namespace); err != nil {
+		logger.Error(err, "Failed to ensure namespace ConfigMaps")
+		if r.Recorder != nil {
+			r.Recorder.Event(rt, corev1.EventTypeWarning, "ConfigMapEnsureError", err.Error())
+		}
+	}
 
 	// 5. Compute config hash from merged configuration (cluster → namespace → CR)
 	configResult, err := ComputeConfigHash(ctx, r.Client, rt.Namespace, &rt.Spec)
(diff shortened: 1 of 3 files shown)


kagenti-operator/internal/controller/agentruntime_controller.go
```diff
@@ -57,16 +57,17 @@ const (
...
+	APIReader client.Reader // uncached reader for cross-namespace ConfigMap reads
 }
 
 // +kubebuilder:rbac:groups=agent.kagenti.dev,resources=agentruntimes,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=agent.kagenti.dev,resources=agentruntimes/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=agent.kagenti.dev,resources=agentruntimes/finalizers,verbs=update
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;update;patch
 // +kubebuilder:rbac:groups=apps,resources=statefulsets,verbs=get;list;watch;update;patch
-// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch
+// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch
```
pdettori, line 70: **suggestion**: RBAC grants `update;patch` on ConfigMaps but the code only uses `Get` and `Create`. Principle of least privilege — consider narrowing to `get;list;watch;create` and adding `update;patch` later if a future change actually needs them.

kagenti-operator/internal/controller/agentruntime_controller.go
```diff
@@ -390,6 +401,73 @@ func (r *AgentRuntimeReconciler) setCondition(rt *agentv1alpha1.AgentRuntime, co
...
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to check ConfigMap %s/%s: %w", namespace, name, err)
+		}
+
+		template := &corev1.ConfigMap{}
+		templateKey := client.ObjectKey{Namespace: ClusterDefaultsNamespace, Name: name}
+		if err := reader.Get(ctx, templateKey, template); err != nil {
+			if apierrors.IsNotFound(err) {
+				logger.V(1).Info("Template ConfigMap not found in kagenti-system, skipping",
+					"name", name)
```
pdettori, line 437: **nit**: Only `template.Data` is copied — `template.BinaryData` is silently dropped. Fine for the current authbridge configs (all text), but adding `BinaryData: template.BinaryData` would be a one-liner that future-proofs against binary content in templates.

pdettori (approved):
Clean, well-scoped fix. The create-if-not-exists + `IsAlreadyExists` guard handles races correctly. Non-fatal error semantics mean the webhook surfaces real issues at pod time without blocking reconciliation. Tests cover all four key cases (create, skip existing, skip missing templates, partial). CI fully green (15/15).

**Areas reviewed**: Go controller logic, RBAC, error handling, tests
**Commits**: 1, DCO signed, correct attribution

PR #316, opened 2026-04-29
feat: add Sandbox workload support to AgentRuntime controller

cwiklik:
## Summary

- Add Sandbox workload support to AgentRuntime controller (unstructured accessor, pod ownership check, RBAC, watch)
- Implement two-phase Sandbox restart for config-driven pod rollout
- Phase 1: scale to 0 + set `kagenti.io/restart-pending` annotation
- Phase 2 (next reconcile): clear `agents.x-k8s.io/pod-name` annotation + remove restart-pending + scale to 1

## Why two-phase restart

The Sandbox controller (kubernetes-sigs/agent-sandbox) tracks its pod via `agents.x-k8s.io/pod-name` annotation. A single-step scale 0→1 causes a race condition where the Sandbox controller enters a stuck error loop because the annotation still points to the deleted pod. The two-phase approach separates scale-down and annotation-clear into different reconciliation windows, avoiding the race with the controller's informer cache.

Upstream issue: kubernetes-sigs/agent-sandbox#581 (pods don't roll out on podTemplate changes)

## Commits

- `3745e25` — initial Sandbox workload support (accessor, ownership check, RBAC, watch)
- `7e22310` — handle Sandbox controller conflicts during restart
- `068ce26` — two-phase restart implementation

## Test plan

- [x] Deploy operator to Kind cluster with Sandbox workload
- [x] Create AgentRuntime with `targetRef.kind: Sandbox`
- [x] Verify config-hash applied to `spec.podTemplate.metadata.annotations`
- [x] Modify watched ConfigMap → verify Phase 1 + Phase 2 in operator logs
- [x] Verify new pod created with updated config
- [x] Verify operator quiescent after restart (no reconcile loop)

Tested end-to-end on Kind cluster with agent-sandbox controller v0.3.10+fix.

Epic: kagenti/kagenti#1155
Companion PR: kagenti/kagenti#1412

### DIFF
--- a/charts/kagenti-operator/templates/rbac/role.yaml
+++ b/charts/kagenti-operator/templates/rbac/role.yaml
@@ -91,6 +91,24 @@ rules:
   - patch
   - update
 - apiGroups:
+  - agents.x-k8s.io
+  resources:
+  - sandboxes
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - agents.x-k8s.io
+  resources:
+  - sandboxes/scale
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
   - apps
   resources:
   - deployments
--- a/kagenti-operator/config/rbac/role.yaml
+++ b/kagenti-operator/config/rbac/role.yaml
@@ -80,6 +80,24 @@ rules:
   - patch
   - update
 - apiGroups:
+  - agents.x-k8s.io
+  resources:
+  - sandboxes
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - agents.x-k8s.io
+  resources:
+  - sandboxes/scale
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
   - apps
   resources:
   - deployments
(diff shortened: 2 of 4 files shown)


huang195 (approved):
Solid implementation of Sandbox workload support with a well-designed two-phase restart. The annotation-based state machine is crash-safe and idempotent — operator crash between phases is handled correctly on the next reconcile. RetryOnConflict usage follows the correct Get-mutate-Update pattern throughout.

**Suggestions (non-blocking):**

1. `"agents.x-k8s.io/pod-name"` is a cross-controller contract — consider extracting to a named constant (e.g., `AnnotationSandboxPodName`) for grep-ability and documentation.

2. In `completeSandboxRestart`, the initial `Get` error returns `(_, false, nil)` for all error types. Transient errors (timeout, network) silently skip the restart check. Consider checking `apierrors.IsNotFound` and only skipping in that case; other errors should propagate.

3. Test coverage for the accessor and ownership logic is good, but the two-phase restart logic (`beginSandboxRestart` / `completeSandboxRestart`) — the most complex part — is untested. Consider adding unit tests with a fake client to exercise both phases and the crash-recovery path.

4. (nit) `sandboxes/scale` subresource permissions are granted but unused — the code uses full-object Update, not the scale subresource endpoint. Harmless but slightly more than minimal.

**Areas reviewed**: Go (controller logic, concurrency, error handling, RBAC), Helm (RBAC template)
**Commits**: 5 commits, all signed-off
**CI status**: all passing (including E2E and integration tests)

PR #324, opened 2026-05-04
fix: use plain strings for Keycloak client attributes in update payload

akram:
## Summary

- `mergeDesiredClientIntoMap` was wrapping attribute values in `[]interface{}` arrays (e.g. `["false"]` instead of `"false"`)
- Keycloak Admin API rejects this with `400: Cannot parse the JSON` on PUT
- This caused operator-managed client registration to silently fail for all workloads when the Keycloak client already existed and needed an update
- Credential secrets were never created, leaving pods stuck in `Init:0/1` waiting for volume mounts

## Root Cause

In `internal/keycloak/admin.go`, `mergeDesiredClientIntoMap` line 414:

```go
// Before (broken):
attrs[k] = []interface{}{v}

// After (fixed):
attrs[k] = v
```

## How it was found

Discovered while testing PR #321 on an OpenShift cluster. The operator authenticated to Keycloak successfully but every client update returned `400 "Cannot parse the JSON"`. Reproduced by manually PUTting the same payload to the Keycloak Admin API — array-wrapped attributes fail, plain string attributes succeed.

## Test plan

- [x] `go test ./internal/keycloak/...` — all 11 tests pass (updated `TestAdmin_RegisterOrFetchClient_updatesDrift`)
- [x] Verified on live OpenShift cluster: client registration succeeds, credential secrets created, pods start

### DIFF
--- a/kagenti-operator/internal/keycloak/admin.go
+++ b/kagenti-operator/internal/keycloak/admin.go
@@ -411,7 +411,7 @@ func mergeDesiredClientIntoMap(m map[string]interface{}, desired *keycloakClient
 		m["attributes"] = attrs
 	}
 	for k, v := range desired.Attributes {
-		attrs[k] = []interface{}{v}
+		attrs[k] = v
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cwiklik (approved):
Clean one-line fix for a real bug — Keycloak rejects array-wrapped attributes with 400. Root cause well-documented, test updated, all CI green.

**Areas reviewed**: Go
**Commits**: 1, signed-off
**CI**: all passing

pdettori (approved):
/lgtm

PR #331, opened 2026-05-05
fix: Update audience scope mapper when included.custom.audience changes

### LINKED ISSUE #330: Audience scope mapper not updated when client ID changes (SPIFFE URI mismatch)
## Bug Description

The `ensureAudienceMapper` function in `internal/keycloak/audience.go` creates audience scope mappers but never updates them if the `included.custom.audience` value changes. This causes a mismatch when clients transition from short-form IDs (`namespace/workload`) to SPIFFE URIs (`spiffe://domain/ns/namespace/sa/workload`).

## Root Cause

`ensureAudienceMapper` uses HTTP POST to create the mapper. If the mapper already exists (HTTP 409 Conflict), it returns `nil` without checking or updating the `included.custom.audience` value:

```go
if resp.StatusCode == http.StatusCreated || resp.StatusCode == http.StatusConflict {
    return nil
}
```

The caller in `clientregistration_controller.go` correctly passes the resolved SPIFFE URI as `AudienceClientID`:

```go
kc.EnsureAudienceScope(ctx, token, keycloak.AudienceParams{
    ClientName:       clientName,    // "namespace/workload"
    AudienceClientID: clientID,      // "spiffe://domain/ns/.../sa/workload"
    ...
})
```

But if the scope was initially created with the short-form audience (e.g., before SPIFFE was enabled, or due to a previous bug), re-running the controller never corrects it.

## Impact

Tokens issued with these audience scopes contain the **short-form** client ID in the `aud` claim (e.g., `redbank-demo-8/redbank-knowledge-agent`), but the actual Keycloak client is registered with the **SPIFFE URI** (e.g., `spiffe://cluster/ns/redbank-demo-8/sa/redbank-knowledge-agent`).

This causes Keycloak token exchange to fail with:
```
access_denied: Client is not within the token audience
```

## Reproduction

1. Deploy workloads without SPIFFE (clients registered as `namespace/workload`)
2. Enable SPIFFE — clients re-register with SPIFFE URIs
3. Operator creates audience scopes with short-form `included.custom.audience`
4. Controller re-reconciles, `EnsureAudienceScope` is called with correct SPIFFE URI
5. POST returns 409, mapper is not updated → stale short-form audience persists

## Proposed Fix

In `ensureAudienceMapper`, on 409 Conflict:
1. GET existing mappers for the scope
2. Check if `included.custom.audience` matches the desired value
3. If not, PUT to update the mapper with the correct audience

Alternatively, the function could always attempt a PUT after a 409 to ensure idempotency.

## Environment

- Keycloak: RHBK 26.4 / Community 26.6
- Operator: latest main
- SPIFFE: enabled with federated-jwt client auth

akram:
## Summary

Fixes #330

When an audience scope mapper already exists (HTTP 409 Conflict), `ensureAudienceMapper` now fetches the existing mapper and compares its `included.custom.audience` value. If it differs from the desired value, the mapper is updated via PUT.

Previously, 409 was treated as success without checking the audience value, causing stale mappers when clients transitioned from short-form IDs (`namespace/workload`) to SPIFFE URIs (`spiffe://domain/ns/namespace/sa/workload`) or vice versa.

## Changes

- Add `ID` field to `protocolMapperRep` (needed for PUT by mapper ID)
- On 409 Conflict in `ensureAudienceMapper`, call new `updateAudienceMapperIfNeeded` to check and fix the audience
- Add `putAudienceMapper` for updating an existing mapper via PUT
- Add two tests: one verifying stale mappers get updated, one verifying correct mappers are left alone

## Root Cause

The `EnsureAudienceScope` caller in `clientregistration_controller.go` correctly passes the resolved SPIFFE URI as `AudienceClientID`:

```go
kc.EnsureAudienceScope(ctx, token, keycloak.AudienceParams{
    ClientName:       clientName,    // "namespace/workload"
    AudienceClientID: clientID,      // "spiffe://domain/ns/.../sa/workload"
})
```

But if the scope was initially created with the short-form audience (e.g., before SPIFFE was enabled), the mapper POST returns 409 and the stale audience was never corrected.

## Test plan

- [x] `TestEnsureAudienceScope_UpdatesStaleMapper` — verifies stale short-form audience is updated to SPIFFE URI via GET + PUT
- [x] `TestEnsureAudienceScope_SkipsUpdateWhenCorrect` — verifies no PUT is issued when audience already matches
- [x] All existing `TestEnsureAudienceScope*` tests pass
- [x] Full keycloak package tests pass (`go test ./internal/keycloak/ -v`)
- [x] Tested on OpenShift cluster (ROSA) with RedBank demo: disabled SPIFFE → verified short-form audiences → re-enabled SPIFFE → operator updated mappers to SPIFFE URIs → token exchange tests pass in both modes

### DIFF
--- a/kagenti-operator/internal/keycloak/audience.go
+++ b/kagenti-operator/internal/keycloak/audience.go
@@ -13,6 +13,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log/slog"
 	"net/http"
 	"net/url"
 	"strings"
@@ -39,7 +40,7 @@ type clientScopeCreateRep struct {
 }
 
 type protocolMapperRep struct {
-	ID              string            `json:"id,omitempty"`
+	ID              string            `json:"id"`
 	Name            string            `json:"name"`
 	Protocol        string            `json:"protocol"`
 	ProtocolMapper  string            `json:"protocolMapper"`
@@ -248,6 +249,9 @@ func (a *Admin) updateAudienceMapperIfNeeded(ctx context.Context, token, realm, 
 		if mappers[i].Name != scopeName || mappers[i].ProtocolMapper != "oidc-audience-mapper" {
 			continue
 		}
+		if mappers[i].Config == nil {
+			continue
+		}
 		if mappers[i].Config["included.custom.audience"] == audience {
 			return nil // already correct
 		}
@@ -255,6 +259,7 @@ func (a *Admin) updateAudienceMapperIfNeeded(ctx context.Context, token, realm, 
 		mappers[i].Config["included.custom.audience"] = audience
 		return a.putAudienceMapper(ctx, token, realm, scopeID, mappers[i])
 	}
+	slog.Debug("no matching audience mapper found for scope", "scope", scopeName, "scopeID", scopeID)
 	return nil
 }
 
--- a/kagenti-operator/internal/keycloak/audience.go
+++ b/kagenti-operator/internal/keycloak/audience.go
@@ -39,6 +39,7 @@ type clientScopeCreateRep struct {
 }
 
 type protocolMapperRep struct {
+	ID              string            `json:"id,omitempty"`
 	Name            string            `json:"name"`
 	Protocol        string            `json:"protocol"`
 	ProtocolMapper  string            `json:"protocolMapper"`
@@ -202,8 +203,12 @@ func (a *Admin) ensureAudienceMapper(ctx context.Context, token, realm, scopeID,
 		return err
 	}
 	defer func() { _ = resp.Body.Close() }()
-	if resp.StatusCode == http.StatusCreated || resp.StatusCode == http.StatusConflict {
+	if resp.StatusCode == http.StatusCreated {
 		return nil
+	}
+	if resp.StatusCode == http.StatusConflict {
+		// Mapper already exists — check if its audience needs updating.
+		return a.updateAudienceMapperIfNeeded(ctx, token, realm, scopeID, scopeName, audience)
 	}
 	body, _ := io.ReadAll(resp.Body)
 	// Mapper may already exist — treat other errors as non-fatal (Python logs and continues).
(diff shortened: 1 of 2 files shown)


huang195:
This seems like a corner case that we have not tested. If I understand it correctly, the scenario is that an agent was initially provisioned without SPIRE enabled, and later, it was restarted with SPIRE enabled. Because error 409 is treated as success, client registration doesn't proceed to update Keycloak with the updated audience scope mapper. Is this right?

I'm curious what is your use case that you needed to restart an agent from no-spire to with-spire. Can't you delete the agent and create a new one?

akram:
That's correct @huang195 . My use case is just that I am running a sequence of tests and I first try token exchange in non-spiffe mode. And then, I enable spiffe and expect token exchange to work as well.

For regular kagentiusers; when they use the kagenti ui; that should correspond to checking unchecking the "Enable SPIFFE" checkbox in the UI. They should certainly hit the same issue.

kagenti-operator/internal/keycloak/audience.go
```diff
@@ -39,6 +39,7 @@ type clientScopeCreateRep struct {
 }
 
 type protocolMapperRep struct {
+	ID              string            `json:"id,omitempty"`
```
pdettori: **suggestion**: The `ID` field uses `json:"id,omitempty"` — consider dropping `omitempty`. The ID is required for `putAudienceMapper` (used to build the PUT endpoint URL). If Keycloak ever returns a mapper without an `id` field, the endpoint would be `.../models/` (empty path segment) — a subtle, silent bug. Using `json:"id"` makes that failure mode explicit at marshal/unmarshal time.

kagenti-operator/internal/keycloak/audience.go
```diff
@@ -213,6 +218,72 @@ func (a *Admin) ensureAudienceMapper(ctx context.Context, token, realm, scopeID,
...
+		return fmt.Errorf("keycloak list mappers decode: %w", err)
+	}
+
+	for i := range mappers {
+		if mappers[i].Name != scopeName || mappers[i].ProtocolMapper != "oidc-audience-mapper" {
+			continue
+		}
+		if mappers[i].Config["included.custom.audience"] == audience {
+			return nil // already correct
+		}
```
pdettori, line 257: **suggestion**: When the loop finishes without finding a matching mapper (no entry with `name == scopeName && protocolMapper == "oidc-audience-mapper"`), the function returns `nil` silently. This could mask issues where the mapper name doesn't match expectations (e.g. scope renamed externally). A debug-level log here (`"no matching audience mapper found for scope %s"`) would help with future troubleshooting.

kagenti-operator/internal/keycloak/audience.go
```diff
@@ -213,6 +218,72 @@ func (a *Admin) ensureAudienceMapper(ctx context.Context, token, realm, scopeID,
...
+		return fmt.Errorf("keycloak list mappers: status %d: %s", resp.StatusCode, truncate(body, 256))
+	}
+
+	var mappers []protocolMapperRep
+	if err := json.Unmarshal(body, &mappers); err != nil {
+		return fmt.Errorf("keycloak list mappers decode: %w", err)
+	}
+
+	for i := range mappers {
+		if mappers[i].Name != scopeName || mappers[i].ProtocolMapper != "oidc-audience-mapper" {
```
pdettori, line 249: **nit**: If `mappers[i].Config` is nil (unlikely with standard Keycloak, but possible with custom providers or future API changes), the map access on this line would panic. A one-line guard `if mappers[i].Config == nil { continue }` would be defensive.

pdettori (approved):
Clean, well-scoped fix for a real production issue (toggling SPIFFE on/off leaves stale audience mappers). The GET+compare+PUT logic on 409 Conflict is the right approach. Tests cover both the update and no-op paths. All CI green.

A few non-blocking suggestions inline.

PR #335, opened 2026-05-07
feat(webhook): Emit per-plugin authbridge config schema

huang195:
## Summary

Follow-up to [kagenti/kagenti-extensions#378](https://github.com/kagenti/kagenti-extensions/pull/378),
which replaces authbridge's top-level `inbound:` / `outbound:` /
`identity:` / `bypass:` / `routes:` YAML blocks with per-plugin config
under `pipeline.*.plugins[].config`. The new authbridge binary rejects
any config whose pipeline is empty (the silent result of yaml.v3
dropping the old top-level keys), so the webhook has to stop
synthesizing the old shape.

**Depends on** #378 — once it merges and `:latest` for
authbridge-envoy / authbridge-light is published, this PR's CI picks
up the new image.

### `pod_mutator.go` — `ensurePerAgentConfigMap` fallback synthesis

The fallback branch that runs when `authbridge-runtime-config` in the
target namespace has no `pipeline:` section is rewritten. Previously
synthesized top-level `inbound:`, `outbound:`, `identity:`, `bypass:`
from `NamespaceConfig`. Now synthesizes a `pipeline:` section with:

- `jwt-validation` (inbound) — config carries `issuer` from
  `NamespaceConfig.Issuer`.
- `token-exchange` (outbound) — config carries `keycloak_url`,
  `keycloak_realm`, `default_policy`, and `identity.type` (mapping
  `federated-jwt` → `spiffe`).

File-path defaults (audience_file, client_id_file, client_secret_file,
jwt_svid_path) are **no longer emitted by the webhook** — the plugin
applies them from its own convention layer at Configure time (see
`authbridge/authlib/plugins/CONVENTIONS.md`). Keeps the webhook
schema-agnostic about filesystem paths and reduces duplication.

When `baseYAML` already has a `pipeline:` section (which the
post-migration Kagenti Helm chart emits), synthesis is skipped; only
`mode` + `listener` overrides layer on top. The chart owns plugin
config contents; the webhook owns per-agent mode selection.

### Tests updated

- `pod_mutator_test.go`:
  - `TestEnsurePerAgentConfigMap_EmptyBaseYAML_FallbackFromNsConfig`
    now asserts the synthesized pipeline via a new `pluginConfigAt`
    helper that navigates `pipeline.<direction>.plugins[<name>].config`.
  - `TestEnsurePerAgentConfigMap_BaseYAML_PreservesExistingFields`
    — baseYAML rewritten to new shape; preservation assertions
    updated accordingly.
  - `TestEnsurePerAgentConfigMap_ListenerOverrides_Merged` —
    baseYAML rewritten; listener assertions unchanged (listener is
    top-level in both schemas).
  - `TestEnsurePerAgentConfigMap_FederatedJWT_MapsToSpiffe` —
    asserts `identity.type = spiffe` under token-exchange config.
    Dropped assertions on default file paths (those moved into
    the plugin).

- `test/e2e/fixtures.go` — two hardcoded `authbridge-runtime-config`
  ConfigMaps rewritten to emit the new per-plugin shape.

### No image tag bumps

The operator references authbridge images as `:latest` throughout
(values.yaml, webhook defaults, e2e suite). Once #378 merges and
publishes `:latest`, operator CI picks it up. Version-pinning and
the atomic authbridge-operator cutover live in the kagenti Helm
chart PR that comes next (tracked separately).

## Test plan

- [x] `go build ./...` clean.
- [x] `go vet ./...` clean.
- [x] All unit tests pass (including the rewritten pod_mutator tests).
- [ ] E2E tests — require a live cluster + the new authbridge image.
  Will run after #378 merges and `:latest` is published.
- [ ] Local kind-cluster smoke test via the Kagenti installer with
  the matched authbridge image — pre-merge gate alongside the
  kagenti Helm chart PR.

## Draft

Draft until #378 lands and the kagenti Helm chart PR is ready —
all three flip from draft to ready together.

### DIFF
--- a/kagenti-operator/internal/webhook/injector/pod_mutator.go
+++ b/kagenti-operator/internal/webhook/injector/pod_mutator.go
@@ -562,6 +562,75 @@ func perAgentConfigMapName(crName string) string {
 	return "authbridge-config-" + crName
 }
 
+// synthesizePipeline builds the per-plugin pipeline section that
+// maps 1:1 to NamespaceConfig (env-var-style authbridge-config
+// values). Used only when the namespace's authbridge-runtime-config
+// ConfigMap has no `pipeline:` of its own — typically a demo or
+// operator-managed namespace that bypassed the Kagenti Helm chart.
+//
+// The synthesized shape matches what plugins expect:
+//   - jwt-validation.config.issuer from NamespaceConfig.Issuer
+//     (rest of the plugin's config comes from its defaults —
+//     audience_file=/shared/client-id.txt, bypass_paths=standard
+//     probes, jwks_url derived from issuer).
+//   - token-exchange.config with Keycloak URL/realm, default_policy,
+//     and identity block keyed off ClientAuthType. File paths fall
+//     through to plugin defaults so operators don't have to
+//     boilerplate them.
+//
+// Empty NamespaceConfig fields are not emitted — the plugin's own
+// defaults apply. That matches the minimum-viable config shown in
+// authbridge/cmd/authbridge/README.md.
+func synthesizePipeline(nsConfig *NamespaceConfig) map[string]interface{} {
+	jwtCfg := map[string]interface{}{}
+	if nsConfig.Issuer != "" {
+		jwtCfg["issuer"] = nsConfig.Issuer
+	}
+
+	tokenCfg := map[string]interface{}{}
+	if nsConfig.KeycloakURL != "" {
+		tokenCfg["keycloak_url"] = nsConfig.KeycloakURL
+	}
+	if nsConfig.KeycloakRealm != "" {
+		tokenCfg["keycloak_realm"] = nsConfig.KeycloakRealm
+	}
+	if nsConfig.DefaultOutboundPolicy != "" {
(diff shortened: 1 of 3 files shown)


pdettori (approved):
Clean migration from flat authbridge config schema to per-plugin `pipeline:` structure. Logic is straightforward and well-tested — if `cfg["pipeline"]` is nil, synthesize from NamespaceConfig; otherwise let the Helm chart's config stand. Bypass-paths default removal is explicitly documented as moved to the plugin's convention layer.

All 16 CI checks pass.

**Two minor suggestions:**

1. `pod_mutator.go` — `synthesizePipeline`: When `ClientAuthType` is non-empty and not `FederatedJWT`, its value is passed through directly as `identity["type"]`. Might be worth a brief comment noting which values the plugin considers valid here — helps future readers understand whether the webhook should validate or just pass through.

2. `pod_mutator_test.go` — `pluginConfigAt`: The local variable `cfg` in the return shadows the outer `cfg` that callers typically have in scope. Consider naming it `pluginCfg` to avoid confusion when reading tests.

PR #336, opened 2026-05-07
feat(webhook): Pass keycloak_url + keycloak_realm to jwt-validation in synthesizePipeline

huang195:
## Summary

Follow-up to kagenti/kagenti-extensions#383 on the operator side. When a namespace's `authbridge-runtime-config` ConfigMap has no `pipeline:` of its own, `synthesizePipeline` builds one from `NamespaceConfig`. Today it passes `keycloak_url` + `keycloak_realm` only to the outbound token-exchange plugin; jwt-validation gets just `issuer`.

kagenti-extensions#383 extends jwt-validation to accept those same two fields and derive `jwks_url` from the INTERNAL Keycloak URL (required so the sidecar can actually reach the JWKS endpoint from inside the cluster — `issuer` is the public hostname, typically unreachable from inside the pod). This PR updates the operator to hand those fields to jwt-validation so synthesized configs work in split-horizon deployments.

## Why we hit this

Observed in kagenti/kagenti#1507 CI (weather-tool-advanced envoy-proxy stderr, with base chart-provided runtime-config — but synthesizePipeline is the fallback path for namespaces that bypass the chart entirely):

```
msg="JWT validation failed" error="fetching JWKS:
  Get \"http://keycloak.localtest.me:8080/realms/kagenti/protocol/openid-connect/certs\":
  dial tcp [::1]:8080: connect: connection refused"
```

The synthesized config emitted only `issuer`. jwt-validation fell back to issuer-derivation, got the public hostname, resolved it to 127.0.0.1 via `localtest.me` wildcard, and hit `connection refused`.

Pre-PR-#378 the binary ran `config.Resolve()` → `deriveJWKSURL()` which derived `inbound.jwks_url` from `outbound.token_url` via a cross-plugin pass. Per-plugin decode gone that pass away; this PR pays that forward by handing the same hints to both plugins.

## Scope

- `pod_mutator.go`: add two lines to `synthesizePipeline` — pass `KeycloakURL` / `KeycloakRealm` into `jwtCfg` alongside the existing pass into `tokenCfg`.
- `pod_mutator_test.go`: extend `TestEnsurePerAgentConfigMap_EmptyBaseYAML_FallbackFromNsConfig` to assert jwt-validation now receives both.

Doc comment on `synthesizePipeline` updated to explain the split-horizon expectation and link to extensions#383.

## Compatibility

Requires the authbridge sidecar binary to accept the new jwt-validation fields (decoded with `DisallowUnknownFields`). The chart in kagenti/kagenti#1507 bumps both pins together — the operator PR and the chart PR gate each other's merge.

## Test plan

- [x] `go test ./internal/webhook/injector/...` — all green (9 ensure-configmap tests, including the updated synthesize assertions)
- [ ] E2E green via the chart PR with matched pins

### DIFF
--- a/kagenti-operator/internal/controller/clientregistration_controller_test.go
+++ b/kagenti-operator/internal/controller/clientregistration_controller_test.go
@@ -425,10 +425,17 @@ func TestClientRegistration_EndToEnd_CredentialsAuthenticate(t *testing.T) {
 		clusterFeatureGatesConfigMap(true),
 		dep,
 		authbridgeConfigMapForTest(clientRegistrationTestNamespace, idp.URL()),
-		keycloakAdminSecretForTest(clientRegistrationTestNamespace),
+		// keycloak-admin-secret lives in the operator namespace (not the
+		// agent namespace) after PR #321's security hardening. The
+		// reconciler reads it from there via OperatorNamespace.
+		keycloakAdminSecretForTest(clientRegistrationTestOperatorNS),
 	).Build()
 
-	r := &ClientRegistrationReconciler{Client: c, Scheme: scheme}
+	r := &ClientRegistrationReconciler{
+		Client:            c,
+		Scheme:            scheme,
+		OperatorNamespace: clientRegistrationTestOperatorNS,
+	}
 	res, err := r.Reconcile(ctx, req)
 	if err != nil || res != (ctrl.Result{}) {
 		t.Fatalf("Reconcile: result=%v err=%v", res, err)
--- a/kagenti-operator/internal/webhook/injector/pod_mutator.go
+++ b/kagenti-operator/internal/webhook/injector/pod_mutator.go
@@ -570,9 +570,16 @@ func perAgentConfigMapName(crName string) string {
 //
 // The synthesized shape matches what plugins expect:
 //   - jwt-validation.config.issuer from NamespaceConfig.Issuer
-//     (rest of the plugin's config comes from its defaults —
+//     (for matching the JWT `iss` claim — the PUBLIC Keycloak URL in
+//     split-horizon deployments). keycloak_url + keycloak_realm are
+//     also passed through so the plugin can derive jwks_url from the
+//     INTERNAL URL — the sidecar actually GETs this URL from inside
+//     the cluster, and the public hostname typically won't resolve
+//     from inside the mesh. See kagenti-extensions#383 for why the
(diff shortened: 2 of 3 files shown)


pdettori (approved):
Clean, focused change. `synthesizePipeline` now passes `keycloak_url` + `keycloak_realm` to jwt-validation with the same guards as token-exchange — both plugins get the internal Keycloak hint needed for split-horizon deployments.

CI: all green except E2E which depends on the paired chart bump in kagenti#1507 (expected).

The bundled test fix for #321's admin-secret namespace mismatch is appreciated — symmetric with the sibling tests and unblocks CI.

pdettori:
## E2E Failure Analysis — False Negative (not caused by this PR)

The E2E failure (`should route HTTP traffic through the injected envoy proxy` + `should auto-create AgentCard with Synced=True`) is a **cross-repo timing issue**, not a regression from this PR's code.

### Root Cause

The `envoy-proxy` container crashloops with:
```
yaml: unmarshal errors:
  line 5: cannot unmarshal !!map into string
  line 10: cannot unmarshal !!map into string
```

This happens because the E2E pulled `authbridge-envoy:latest` **after** the `v0.5.0-rc.1` tag was published (at 23:21Z), which includes extensions #378's config refactoring. The rc.1 binary changed how plugin `config` fields are decoded — it now expects string-typed config, but the operator generates nested YAML maps.

### Timeline

| Time (UTC) | Event |
|---|---|
| 19:57 | Last Build-Publish on extensions `main` → pushes `:latest` |
| 22:48 | Main CI E2E **passes** (pulls the old `:latest`) |
| 23:21 | `v0.5.0-rc.1` Build-Publish → **overwrites** `:latest` with rc.1 code |
| 23:47 | This PR's E2E pulls `:latest` → gets the rc.1 image → 💥 |
| 01:27 | Extensions #383 merges to main → new `:latest` fixes the parsing |

### Evidence

- The PR only changes `synthesizePipeline`, but the E2E uses the **legacy ValueFrom config path** (confirmed by controller logs: `"Using legacy ValueFrom config path"`). `synthesizePipeline` is never called.
- Main's E2E passed at 22:48Z with the same operator code (commit `4a7d291`), before the rc.1 image was published.
- Unit Tests + Integration Tests pass on this PR.

### Fix

Re-running CI should resolve this — `authbridge-envoy:latest` now includes both #378 and #383 (merged at 01:27Z), which properly handles nested-map config fields.

PR #350, opened 2026-05-11
fix(keycloak): propagate and verify audience mapper errors (#348)

### LINKED ISSUE #348: bug: ensureAudienceMapper silently swallows errors, leaving scopes without mappers
## Problem

When the `clientregistration` controller registers multiple agents concurrently, `ensureAudienceMapper()` can fail transiently (e.g., Keycloak rate-limiting, scope not yet visible after creation). The error is **silently discarded**, leaving the audience scope without a protocol mapper.

User tokens then lack the required `aud` claim for that agent, and AuthBridge returns 401:

```
JWT validation failed — "aud" not satisfied
```

## Root Cause

In [`audience.go`](https://github.com/kagenti/kagenti-operator/blob/main/kagenti-operator/internal/keycloak/audience.go), both call sites discard the error:

```go
// Line 90 — scope already existed
_ = a.ensureAudienceMapper(ctx, token, realm, scopeID, scopeName, audience)

// Line 108 — scope just created
_ = a.ensureAudienceMapper(ctx, token, realm, scopeID, scopeName, audience)
```

If the mapper POST fails for any reason other than 409 Conflict (e.g., 500, network timeout, Keycloak not yet consistent), the scope is returned without a mapper and the controller considers registration complete.

## Observed Behavior

On a Kind cluster with two agents registered in the same reconcile cycle:

| Scope | Mapper | Token `aud` |
|-------|--------|-------------|
| `agent-team1-weather-service-2-aud` | `oidc-audience-mapper` with `spiffe://...weather-service-2` | Correct |
| `agent-team1-weather-service-aud` | **empty — no mapper** | Missing → 401 |

## Not Fixed By

PR #331 addressed the case where a mapper exists but has a stale `included.custom.audience`. It does not cover the case where no mapper was created at all.

## Suggested Fix

1. **Propagate the error** from `ensureAudienceMapper` so the controller requeues:

```go
if err := a.ensureAudienceMapper(ctx, token, realm, scopeID, scopeName, audience); err != nil {
    return "", fmt.Errorf("audience scope %q created but mapper failed: %w", scopeName, err)
}
```

2. Optionally add a **reconcile-time check** that verifies the mapper exists on the scope, and re-creates it if missing (defense in depth against Keycloak inconsistency).

## Related

- kagenti/kagenti#1526 — end-to-end bug report (this is bug 2 of 2)
- #330 / #331 — audience mapper update fix (partial, does not cover this case)

Assisted-By: Claude Code

cwiklik:
## Summary

- Propagate `ensureAudienceMapper` errors instead of silently discarding them with `_ =`, so the controller sees failures and can act on them
- Add `verifyAudienceMapper` defense-in-depth check that runs on every reconcile — GETs the scope's protocol mappers and re-creates the `oidc-audience-mapper` if missing (catches scopes broken by prior transient failures)
- Make `updateAudienceMapperIfNeeded` return an error when no matching mapper is found (was a silent no-op)

Fixes #348

## Changes

| File | What |
|------|------|
| `internal/keycloak/audience.go` | Propagate errors on both code paths (existing + new scope); add `verifyAudienceMapper`; remove unused `slog` import |
| `internal/keycloak/audience_test.go` | Add `MapperFailurePropagated` test (mapper 500 → error surfaces); add `VerifyRecreatesMissingMapper` test (missing mapper detected and re-created); update existing tests for verify GET call |

### Unit tests — all 6 pass

```
=== RUN   TestEnsureAudienceScope
--- PASS: TestEnsureAudienceScope (0.00s)
=== RUN   TestEnsureAudienceScope_UpdatesStaleMapper
--- PASS: TestEnsureAudienceScope_UpdatesStaleMapper (0.00s)
=== RUN   TestEnsureAudienceScope_SkipsUpdateWhenCorrect
--- PASS: TestEnsureAudienceScope_SkipsUpdateWhenCorrect (0.00s)
=== RUN   TestEnsureAudienceScope_MapperFailurePropagated
--- PASS: TestEnsureAudienceScope_MapperFailurePropagated (0.00s)
=== RUN   TestEnsureAudienceScope_VerifyRecreatesMissingMapper
--- PASS: TestEnsureAudienceScope_VerifyRecreatesMissingMapper (0.00s)
=== RUN   TestEnsureAudienceScope_Disabled
--- PASS: TestEnsureAudienceScope_Disabled (0.00s)
PASS
ok  	github.com/kagenti/operator/internal/keycloak	1.051s
```

### Kind cluster — local image build + deploy

Built operator image from this branch, loaded into Kind cluster via `ctr import`, rolled out `kagenti-controller-manager` deployment with `imagePullPolicy: Never`.

```
$ kubectl get pods -n kagenti-system -l control-plane=controller-manager
NAME                                          READY   STATUS    RESTARTS   AGE
kagenti-controller-manager-74d4cb5446-9nvt2   1/1     Running   0          27s
```

### Kind cluster — new agent deploy validates fix

Deployed a new agent (`weather-service-777`) in `team1` namespace via Sandbox CR with source build. The operator reconciled the new pod and created the audience scope with the correct mapper:

```
=== Audience scopes ===
  agent-team1-weather-service-777-aud  (id: 5d6c7190-c73e-41b4-b35b-b928bb5b9989)
  agent-team1-weather-service-aud      (id: 311fb6d9-40ec-4c9f-adf6-65704969f44f)

=== weather-service-777 mapper ===
  mapper: agent-team1-weather-service-777-aud
  type:   oidc-audience-mapper
  audience: spiffe://localtest.me/ns/team1/sa/weather-service-777
```

Both agents have correct audience scopes with mappers — no silent failures, no missing mappers.

Assisted-By: Claude Code

### DIFF
--- a/kagenti-operator/internal/keycloak/audience.go
+++ b/kagenti-operator/internal/keycloak/audience.go
@@ -87,7 +87,9 @@ func (a *Admin) getOrCreateAudienceClientScope(ctx context.Context, token, realm
 		return "", err
 	}
 	if scopeID != "" {
-		_ = a.ensureAudienceMapper(ctx, token, realm, scopeID, scopeName, audience)
+		if err := a.ensureAudienceMapper(ctx, token, realm, scopeID, scopeName, audience); err != nil {
+			return "", fmt.Errorf("ensure audience mapper for existing scope %q: %w", scopeName, err)
+		}
 		return scopeID, nil
 	}
 
@@ -105,7 +107,9 @@ func (a *Admin) getOrCreateAudienceClientScope(ctx context.Context, token, realm
 	if scopeID == "" {
 		return "", fmt.Errorf("create client scope %q returned empty id", scopeName)
 	}
-	_ = a.ensureAudienceMapper(ctx, token, realm, scopeID, scopeName, audience)
+	if err := a.ensureAudienceMapper(ctx, token, realm, scopeID, scopeName, audience); err != nil {
+		return "", fmt.Errorf("ensure audience mapper for new scope %q: %w", scopeName, err)
+	}
 	return scopeID, nil
 }
 
--- a/kagenti-operator/internal/keycloak/audience.go
+++ b/kagenti-operator/internal/keycloak/audience.go
@@ -225,30 +225,41 @@ func (a *Admin) ensureAudienceMapper(ctx context.Context, token, realm, scopeID,
 	return nil
 }
 
-// updateAudienceMapperIfNeeded fetches the existing mapper for the scope and updates
-// its included.custom.audience if it differs from the desired value.
-func (a *Admin) updateAudienceMapperIfNeeded(ctx context.Context, token, realm, scopeID, scopeName, audience string) error {
+// listAudienceMappers fetches all protocol mappers for a client scope.
+func (a *Admin) listAudienceMappers(ctx context.Context, token, realm, scopeID string) ([]protocolMapperRep, error) {
 	base := trimBaseURL(a.BaseURL)
 	endpoint := base + "/admin/realms/" + url.PathEscape(realm) + "/client-scopes/" + url.PathEscape(scopeID) + "/protocol-mappers/models"
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, endpoint, nil)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	req.Header.Set("Authorization", "Bearer "+token)
 
 	resp, err := a.httpc().Do(req)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	defer func() { _ = resp.Body.Close() }()
 	body, _ := io.ReadAll(resp.Body)
 	if resp.StatusCode != http.StatusOK {
-		return fmt.Errorf("keycloak list mappers: status %d: %s", resp.StatusCode, truncate(body, 256))
+		return nil, fmt.Errorf("keycloak list mappers: status %d: %s", resp.StatusCode, truncate(body, 256))
 	}
 
 	var mappers []protocolMapperRep
 	if err := json.Unmarshal(body, &mappers); err != nil {
-		return fmt.Errorf("keycloak list mappers decode: %w", err)
+		return nil, fmt.Errorf("keycloak list mappers decode: %w", err)
+	}
+	return mappers, nil
+}
+
+// updateAudienceMapperIfNeeded fetches the existing mapper for the scope and updates
+// its included.custom.audience if it differs from the desired value.
+// Returns an error if no matching mapper is found — this treats "no match" as a real
+// failure (e.g. Keycloak race or name mismatch) rather than silently ignoring it.
+func (a *Admin) updateAudienceMapperIfNeeded(ctx context.Context, token, realm, scopeID, scopeName, audience string) error {
+	mappers, err := a.listAudienceMappers(ctx, token, realm, scopeID)
+	if err != nil {
+		return err
 	}
 
 	for i := range mappers {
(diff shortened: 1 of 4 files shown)


cwiklik:
## CI Note: Unit Test failure is pre-existing, not related to this PR

The failing test is `mapWorkloadToAgentCards` in `indexers_test.go:115` — a Sandbox→AgentCard mapping test that errors with `field label not supported: .spec.targetRef.name`. This is unrelated to the keycloak audience mapper changes in this PR.

Main branch CI (`ci.yaml`) also shows this failure:

| Run | Branch | Conclusion |
|-----|--------|------------|
| 25571428211 | main | failure |
| 25571405507 | main | failure |
| 25569991617 | main | success |

All 6 keycloak audience tests pass (including 2 new ones added by this PR).

kagenti-operator/internal/keycloak/audience.go
```diff
@@ -289,6 +294,50 @@ func (a *Admin) putAudienceMapper(ctx context.Context, token, realm, scopeID str
...
+
+	for i := range mappers {
+		if mappers[i].Name != scopeName || mappers[i].ProtocolMapper != "oidc-audience-mapper" {
+			continue
+		}
+		if mappers[i].Config != nil && mappers[i].Config["included.custom.audience"] == audience {
+			return nil
+		}
+		if mappers[i].Config == nil {
+			mappers[i].Config = make(map[string]string)
```
rubambiza, line 327: suggestion: `verifyAudienceMapper` and `updateAudienceMapperIfNeeded` share ~25 lines of GET-and-parse logic (endpoint, auth header, read body, status check, unmarshal, match loop). Consider extracting a `listAudienceMappers(ctx, token, realm, scopeID) ([]protocolMapperRep, error)` helper — both callers would benefit and the two match-loop behaviors (update-only vs create-if-missing) become clearer. Not blocking.

kagenti-operator/internal/keycloak/audience.go
```diff
@@ -259,8 +265,7 @@ func (a *Admin) updateAudienceMapperIfNeeded(ctx context.Context, token, realm,
 		mappers[i].Config["included.custom.audience"] = audience
 		return a.putAudienceMapper(ctx, token, realm, scopeID, mappers[i])
 	}
-	slog.Debug("no matching audience mapper found for scope", "scope", scopeName, "scopeID", scopeID)
-	return nil
+	return fmt.Errorf("no matching audience mapper found for scope %q (scopeID %s)", scopeName, scopeID)
```
rubambiza, line 278: suggestion: `updateAudienceMapperIfNeeded` now returns an error when no matching mapper is found. This changes behavior for the 409 Conflict path in `ensureAudienceMapper`: a 409 with no name match would previously be silently ignored, now it errors. This is the right direction for visibility, but worth a brief comment on the function (or in the error string) noting that "no match" is now treated as a real failure (e.g. case mismatch or Keycloak race) rather than the previous silent no-op — makes the behavior change easier to spot later.

kagenti-operator/internal/keycloak/audience.go
```diff
@@ -289,6 +294,50 @@ func (a *Admin) putAudienceMapper(ctx context.Context, token, realm, scopeID str
 	return fmt.Errorf("keycloak update audience mapper: status %d: %s", resp.StatusCode, truncate(body, 256))
 }
 
+// verifyAudienceMapper is a defense-in-depth check that runs on every reconcile.
```
rubambiza, line 307: nit: `verifyAudienceMapper` adds an extra `GET /protocol-mappers/models` on every reconcile for every audience-enabled scope. The defense-in-depth tradeoff is reasonable, but worth acknowledging the cost in the doc comment (e.g. "one extra GET per reconcile; accepted cost for catching scopes left broken by prior transient failures").

rubambiza (approved):
## Review Summary

Well-scoped fix for #348: replaces error-discarding `_ =` calls with proper propagation, adds a defense-in-depth `verifyAudienceMapper` that re-creates missing mappers on reconcile, and adds regression tests for both failure propagation and mapper re-creation. Tests look solid and all 15 CI checks pass.

**Areas reviewed**: Go (internal/keycloak)
**Commits**: 3, all DCO signed-off, conventional format, `Assisted-By:` per project policy
**CI**: all 15 checks green

Approving. Left a couple of non-blocking suggestions/nits inline.

Assisted-By: Claude Code

PR #357, opened 2026-05-13
Docs: provide text for 'oc explain' on operator CRDs

esnible:
## Summary

Kubernetes and OpenShift admins use `oc explain` to understand CRDs.

The previous output for our CRDs was based on generated Go code and very confusing:

```
oc explain agentcard
...
DESCRIPTION:
     AgentCard is the Schema for the agentcards API.
...
```

This PR gives a real description for each of our CRDs.

```bash
make manifests
kubectl apply -f config/crd/bases/agent.kagenti.dev_agentcards.yaml 
kubectl apply -f config/crd/bases/agent.kagenti.dev_agentruntimes.yaml
oc explain agentcard
oc explain agentruntimes
```

With this change, the output should include

```
DESCRIPTION:
     AgentCard binds an A2A agent card to a backing workload. The controller
     periodically fetches the card from the referenced workload, verifies its
     JWS signature and SPIFFE identity against the configured trust domain, and
     caches the result in status.
```

and

```
DESCRIPTION:
     AgentRuntime attaches runtime configuration to a backing workload
     classified as an agent or tool, providing per-workload overrides for SPIFFE
     identity and OpenTelemetry tracing. The controller reports pod
     configuration coverage and phase in status.
```

Note: Running `make manifest`, overrides The namespaceSelector, objectSelector, and timeoutSeconds: 10 that were deliberately added manually to _config/webhook/manifests.yaml_ in commits 19918ea and b89651f .  If we plan to continue allowing `make manifest` to work, we should probably also make the changes to _authbridge_webhook.go_:214.

### DIFF
--- a/kagenti-operator/api/v1alpha1/agentcard_types.go
+++ b/kagenti-operator/api/v1alpha1/agentcard_types.go
@@ -345,7 +345,9 @@ type SkillParameter struct {
 // +kubebuilder:printcolumn:name="LastSync",type="date",JSONPath=".status.lastSyncTime",description="Last Sync Time"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 
-// AgentCard is the Schema for the agentcards API.
+// AgentCard binds an A2A agent card to a backing workload. The controller periodically
+// fetches the card from the referenced workload, verifies its JWS signature and SPIFFE
+// identity against the configured trust domain, and caches the result in status.
 type AgentCard struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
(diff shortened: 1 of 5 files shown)


huang195 (approved):
### Summary

Swaps the boilerplate `"X is the Schema for the X API"` kubebuilder descriptions for real, user-facing ones on `AgentCard` and `AgentRuntime`; the regenerated CRD YAMLs pick up the same text. Claims check out against the controllers — JWS signature + SPIFFE trust-domain verification on the AgentCard side, phase + per-workload SPIRE/OTel overrides on the AgentRuntime side.

**Areas reviewed**: Go API types, CRD YAML
**Commits**: 2, both signed-off (the second restores the manual webhook-manifests override that `make manifests` clobbered in the first)
**CI status**: all green

### Follow-up (not blocking)

PR body correctly calls out that `make manifests` overwrites the manual `namespaceSelector` / `objectSelector` / `timeoutSeconds` in `config/webhook/manifests.yaml` (restored here via `afe62a9`) and suggests baking those into `authbridge_webhook.go:214`. Worth a tracking issue so the next person running `make manifests` doesn't silently re-regress the webhook config — happy to file one after this merges if it's not already tracked.

PR #360, opened 2026-05-14
fix(keycloak): stabilize audience mapper registration for existing scopes

### LINKED ISSUE #358: fix(keycloak): verifyAudienceMapper defense-in-depth never runs due to early return in getOrCreateAudienceClientScope

pdettori:
## Summary

- **Match Python AuthBridge sidecar pattern**: for existing scopes, never POST a new mapper — only verify via GET and update via PUT
- **Prevent ghost-409 cascades**: repeated mapper POSTs on every reconcile corrupted Keycloak's internal name index, requiring Keycloak restarts to recover
- **Add `createAudienceMapperBestEffort`**: on 409, verifies mapper state via GET before returning success; gracefully handles Keycloak ghost-conflicts

## Root Cause

The operator called `ensureAudienceMapper` (a POST) on every reconcile for existing scopes. The Python AuthBridge sidecar only POSTs a mapper when creating a brand-new scope. Repeated POSTs trigger 409 conflicts, and if a mapper ever enters a corrupted state, the 409 cascades into ghost-conflict entries in Keycloak's DB that persist across pod restarts.

## Test plan

- [x] Unit tests pass (`go test ./internal/keycloak/ -run TestEnsureAudienceScope`)
- [x] Built and deployed to Kind cluster (`localhost/kagenti-operator:fix-358`)
- [x] Verified stable re-registration: delete/recreate AgentRuntime → no 409 errors
- [x] Verified concurrent reconciles: rapid back-to-back deployment restarts → all succeed cleanly
- [x] Verified mapper integrity: correct `oidc-audience-mapper` with SPIFFE audience after all operations
- [x] Verified both agents (weather-service, a2a-contact-extractor) register cleanly in parallel

Fixes #358

Assisted-By: Claude Code

### DIFF
--- a/kagenti-operator/internal/keycloak/audience.go
+++ b/kagenti-operator/internal/keycloak/audience.go
@@ -47,6 +47,8 @@ type protocolMapperRep struct {
 	Config          map[string]string `json:"config"`
 }
 
+const oidcAudienceMapper = "oidc-audience-mapper"
+
 // AudienceScopeName derives the realm client-scope name from CLIENT_NAME (same as Python).
 func AudienceScopeName(clientName string) string {
 	return "agent-" + strings.ReplaceAll(clientName, "/", "-") + "-aud"
@@ -186,7 +188,7 @@ func (a *Admin) ensureAudienceMapper(ctx context.Context, token, realm, scopeID,
 	mapper := protocolMapperRep{
 		Name:            scopeName,
 		Protocol:        "openid-connect",
-		ProtocolMapper:  "oidc-audience-mapper",
+		ProtocolMapper:  oidcAudienceMapper,
 		ConsentRequired: false,
 		Config: map[string]string{
 			"included.custom.audience": audience,
@@ -271,7 +273,7 @@ func (a *Admin) updateAudienceMapperIfNeeded(ctx context.Context, token, realm, 
 		if mappers[i].Name != scopeName {
 			continue
 		}
-		if mappers[i].ProtocolMapper != "oidc-audience-mapper" {
+		if mappers[i].ProtocolMapper != oidcAudienceMapper {
 			if err := a.deleteMapper(ctx, token, realm, scopeID, mappers[i].ID); err != nil {
 				return fmt.Errorf("delete stale mapper %q (type %q) for scope %q: %w",
 					mappers[i].Name, mappers[i].ProtocolMapper, scopeName, err)
@@ -321,7 +323,7 @@ func (a *Admin) createAudienceMapperBestEffort(ctx context.Context, token, realm
 	mapper := protocolMapperRep{
 		Name:            scopeName,
 		Protocol:        "openid-connect",
-		ProtocolMapper:  "oidc-audience-mapper",
+		ProtocolMapper:  oidcAudienceMapper,
 		ConsentRequired: false,
 		Config: map[string]string{
 			"included.custom.audience": audience,
@@ -360,7 +362,7 @@ func (a *Admin) createAudienceMapperBestEffort(ctx context.Context, token, realm
 			return nil
 		}
 		for i := range mappers {
-			if mappers[i].Name == scopeName && mappers[i].ProtocolMapper == "oidc-audience-mapper" {
+			if mappers[i].Name == scopeName && mappers[i].ProtocolMapper == oidcAudienceMapper {
 				if mappers[i].Config != nil && mappers[i].Config["included.custom.audience"] == audience {
 					return nil
 				}
@@ -420,7 +422,7 @@ func (a *Admin) verifyAudienceMapper(ctx context.Context, token, realm, scopeID,
 		if mappers[i].Name != scopeName {
 			continue
 		}
-		if mappers[i].ProtocolMapper != "oidc-audience-mapper" {
+		if mappers[i].ProtocolMapper != oidcAudienceMapper {
 			if err := a.deleteMapper(ctx, token, realm, scopeID, mappers[i].ID); err != nil {
 				return fmt.Errorf("delete stale mapper %q (type %q): %w",
 					mappers[i].Name, mappers[i].ProtocolMapper, err)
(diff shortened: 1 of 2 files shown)


kagenti-operator/internal/keycloak/audience.go
```diff
@@ -254,28 +259,126 @@ func (a *Admin) listAudienceMappers(ctx context.Context, token, realm, scopeID s
...
+					mappers[i].Config = make(map[string]string)
+				}
+				mappers[i].Config["included.custom.audience"] = audience
+				return a.putAudienceMapper(ctx, token, realm, scopeID, mappers[i])
+			}
+		}
+		// Ghost-409: mapper not visible. Return nil; next reconcile will retry.
+		return nil
+	}
+	// Other errors: non-fatal, reconciler retries.
```
cwiklik: suggestion: `createAudienceMapperBestEffort` returns `nil` for all non-409/non-2xx status codes here. This silently swallows 500s, 503s, auth errors, etc. While `verifyAudienceMapper` will re-check on the next reconcile, returning the error here would let the controller re-queue immediately rather than waiting for the next periodic pass.

Consider:
```go
body, _ := io.ReadAll(resp.Body)
return fmt.Errorf("keycloak create mapper best-effort: status %d: %s", resp.StatusCode, truncate(body, 256))
```

Non-blocking — the current behavior is safe given the reconcile loop, just slower to recover from persistent Keycloak errors.

kagenti-operator/internal/keycloak/audience.go
```diff
@@ -110,7 +112,10 @@ func (a *Admin) getOrCreateAudienceClientScope(ctx context.Context, token, realm
 		return "", fmt.Errorf("create client scope %q returned empty id", scopeName)
 	}
 	if err := a.ensureAudienceMapper(ctx, token, realm, scopeID, scopeName, audience); err != nil {
-		return "", fmt.Errorf("ensure audience mapper for new scope %q: %w", scopeName, err)
+		// Mapper creation failed for a brand-new scope. This can happen if createClientScope
+		// hit a 409 race and another reconcile already created the mapper. Non-fatal —
+		// verifyAudienceMapper will repair on next pass.
```
cwiklik: nit: Comment says "will repair on next pass" but `verifyAudienceMapper` runs later in the *same* `EnsureAudienceScope` call (not the next reconcile). Suggest: "Non-fatal — verifyAudienceMapper will repair below in this reconcile."

cwiklik (approved):
Clean fix that correctly aligns the operator with the Python AuthBridge sidecar pattern: only POST mappers for brand-new scopes, verify via GET+PUT for existing ones. The `createAudienceMapperBestEffort` + ghost-409 handling is sound. Good test coverage for the corrupted-mapper and repair scenarios.

All 15 CI checks pass. 2 commits, both DCO-signed.

PR #366, opened 2026-05-20
feat(controller): scope MLflow RBAC to per-agent namespaced Role

rh-dnagornuks:
## Summary
The MLflow controller previously bound each agent ServiceAccount to the broad `mlflow-operator-mlflow-integration` ClusterRole, granting access to all MLflow experiments, registered models, and gateway endpoints cluster-wide. This PR replaces that with a least-privilege namespaced Role per agent Deployment, scoped via `resourceNames` to only the agent's own experiment, and a corresponding RoleBinding — both owned by the Deployment so they are garbage-collected on deletion.

## Changes
- `mlflow_controller.go` — replace `ensureRoleBinding` (ClusterRole binding) with `ensureScopedRBAC`, which creates a namespaced Role granting only `get`+`update` on the specific `mlflowexperiments` resource name, plus a RoleBinding to the agent SA; remove the now-unused `MLflowClusterRole` field and `clusterRoleName()` helper
- `mlflow_controller.go` — add `Owns(&rbacv1.Role{})` to the controller watch so Role deletions trigger reconciliation
- `config/rbac/role.yaml` and `charts/.../rbac/role.yaml` — add `mlflowexperiments` (get/list/watch/update) and `roles` (create/get/list/watch/update) RBAC rules, regenerated from kubebuilder markers
- `docs/mlflow-integration.md` — update Authentication section to document the new two-resource scoped RBAC pattern
- `mlflow_controller_test.go` — update tests to assert scoped Role creation with correct `resourceNames`, update RoleBinding assertions to check `Kind: Role`, remove obsolete `clusterRoleName` helper tests

## Testing
- [x] Unit tests updated and passing (`make test`)
- [x] Linter passing (`make lint`)
- [x] Manually verified Role/RoleBinding created with correct `resourceNames` scoping on an OpenShift cluster

## Related Issues

Resolves: [RHAIENG-4446](https://redhat.atlassian.net/browse/RHAIENG-4446)

### DIFF
--- a/kagenti-operator/internal/controller/mlflow_controller.go
+++ b/kagenti-operator/internal/controller/mlflow_controller.go
@@ -40,9 +40,13 @@ import (
 )
 
 const (
-	// DefaultMLflowClusterRole is the ClusterRole managed by the MLflow operator
-	// for agent access to MLflow resources (RHOAI 3.4+).
-	DefaultMLflowClusterRole = "mlflow-operator-mlflow-integration"
+	// MLflowExperimentsAPIGroup is the API group used by the MLflow Kubernetes
+	// authorization plugin for SubjectAccessReview checks.
+	MLflowExperimentsAPIGroup = "mlflow.opendatahub.io"
+
+	// MLflowExperimentsResource is the virtual resource the MLflow gateway checks
+	// when authorizing experiment-level operations.
+	MLflowExperimentsResource = "mlflowexperiments"
 
 	// MLflow annotation keys stored on the PodTemplateSpec.
 	AnnotationMLflowExperimentID   = "mlflow.kagenti.io/experiment-id"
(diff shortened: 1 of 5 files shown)


kagenti-operator/internal/controller/mlflow_controller.go
```diff
@@ -40,9 +40,13 @@ import (
 )
```
pdettori, line 41: nit: The kubebuilder marker adds `get;list;watch;update` for `mlflowexperiments` on the operator's ClusterRole. Since the controller only verifies experiment existence for `resourceNames` scoping, `list` and `watch` may not be strictly necessary (just `get` might suffice). Not blocking — broader permissions are fine for controller reliability and future use.

pdettori (approved):
Well-structured least-privilege RBAC refactor. Replaces a broad ClusterRoleBinding (all agents access all MLflow resources cluster-wide) with namespace-scoped Roles locked to each agent's specific experiment via `resourceNames`. Both Role and RoleBinding are owned by the Deployment for automatic GC, and the controller watches Role deletions for re-reconciliation.

**Areas reviewed**: Go controller logic, Helm/K8s YAML (RBAC parity kustomize ↔ chart), tests, docs, security
**Commits**: 1 commit, signed-off ✓, proper `Assisted-By` trailer
**CI**: All 16 checks passing

PR #370, opened 2026-05-21
Wire AllowedAudiences from AgentRuntime CRD to AuthProxy sidecar

akram:
## Summary

- Adds `spec.identity.allowedAudiences` field to the AgentRuntime CRD, allowing per-agent JWT audience overrides
- Wires the field through `AgentRuntimeOverrides` → `ResolvedConfig` → per-agent ConfigMap generation
- Injects `allowed_audiences` into the `pipeline.inbound.plugins[jwt-validation].config` section of the per-agent ConfigMap
- Transitional mechanism to support application-to-agent flows (e.g. playground → agent) until the auth model is finalized

### Example usage

```yaml
apiVersion: agent.kagenti.dev/v1alpha1
kind: AgentRuntime
metadata:
  name: my-agent-runtime
spec:
  type: agent
  targetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: my-agent
  identity:
    allowedAudiences:
      - playground
      - kagenti-agents
```

### Changed files

| File | Change |
|------|--------|
| `api/v1alpha1/agentruntime_types.go` | Add `AllowedAudiences []string` to `IdentitySpec` |
| `internal/webhook/injector/agentruntime_config.go` | Extract from CR, add to `AgentRuntimeOverrides` |
| `internal/webhook/injector/resolved_config.go` | Add to `ResolvedConfig`, merge with AR precedence |
| `internal/webhook/injector/pod_mutator.go` | `injectAllowedAudiences()` helper, pass through both call sites |
| `internal/webhook/injector/pod_mutator_test.go` | Tests for injection and nil cases |
| `api/v1alpha1/zz_generated.deepcopy.go` | Regenerated |
| `config/crd/bases/agent.kagenti.dev_agentruntimes.yaml` | Regenerated CRD manifest |

Closes #368

## Test plan

- [x] Unit tests pass (`go test ./internal/webhook/injector/...`)
- [x] Verified on OpenShift (ROSA) with `redbank-demo-test-envoy` namespace:
  - Patched AgentRuntime CRs with `allowedAudiences: ["redbank-demo-test-envoy", "account"]`
  - Confirmed `allowed_audiences` appears in per-agent ConfigMaps under `pipeline.inbound.plugins[jwt-validation].config`
  - Confirmed authbridge sidecar starts and accepts user tokens with matching audiences (`inbound authorized`)
  - Full E2E: playground → orchestrator → banking/knowledge agents working with envoy-sidecar mode
[...]

PR #369, opened 2026-05-20
feat(authbridge): Wire mtlsMode through to per-agent config

huang195:
## Summary

Wires the now-merged [kagenti-extensions PR #424](https://github.com/kagenti/kagenti-extensions/pull/424) (mTLS between authbridge sidecars on the proxy-sidecar / lite paths) through the operator so it can be enabled declaratively per workload — and folds in two pre-existing tech-debt items along the way.

The mTLS feature is already fully implemented on the authbridge side: the byte-peek listener, X.509 SVID consumption, permissive-with-fallback / strict modes — all live in the merged PR. This PR makes the operator render the right config, reject impossible combos, and closes two related rollout / chart-sync gaps.

### mtlsMode wiring (commit 1)

* **`AgentRuntimeSpec.MTLSMode`** — new enum field (`disabled` | `permissive` | `strict`), default `disabled`. Resolution chain mirrors `AuthBridgeMode`: CR > namespace `authbridge-runtime-config` `mtls.mode` > default.

* **Per-agent ConfigMap renders `mtls: {mode: <value>}`** when resolved mode is non-disabled. Cert paths intentionally omitted — they match authbridge's bundled-spiffe-helper defaults (`/opt/svid.pem` etc.). Stale `mtls` blocks scrubbed when toggling back to disabled.

* **`mtlsMode != disabled` auto-enables SPIRE** for the workload (sets `spireEnabled` for that injection + ensures the dedicated ServiceAccount). Operators don't need to add the `kagenti.io/spiffe-helper-inject` label separately. We don't mutate the namespace `authbridge-config` ConfigMap — that would race with the kagenti meta-chart for ownership.

* **Validating webhook rejects `mtlsMode != disabled` + `authBridgeMode = envoy-sidecar`.** Envoy SDS isn't currently configured by the kagenti envoy-config. Accepting that combo would silently run plaintext while the user believed they had strict mTLS — exactly the wrong failure mode for a security-relevant posture claim. Error message names the supported modes (`proxy-sidecar` / `lite`) and explicitly flags this as a current limitation, not a permanent one.

* **Spec changes to `AuthBridgeMode` and `MTLSMode` feed the resolved-config hash.** Editing either CR field re-stamps `kagenti.io/config-hash` on the pod template, so the Deployment rolls automatically. (Bonus: this fixes a pre-existing gap where `AuthBridgeMode` changes weren't triggering rollouts.)

### Namespace-CM-edit rollout gap (commit 2)

[...]

PR #374, opened 2026-05-23
ci(release): parallelize image builds + add buildx cache + fix e2e

huang195:
## What

Three changes to `.github/workflows/release.yml`, all addressing pre-existing CI issues.

### 1. Parallelize the two image builds (matrix)

`build-and-push` was a single job building `kagenti-operator` and `agentcard-signer` sequentially. With QEMU-emulated `linux/arm64` builds, each took ~20 minutes. Total wall time: ~42 minutes per run.

Refactored into a matrix of two jobs running in parallel on separate runners. Wall time drops to ~22 minutes (the longest of the two parallel builds).

### 2. Add buildx GitHub Actions cache

Each matrix instance now has `cache-from: type=gha` / `cache-to: type=gha,mode=max` with a per-image `scope`. Warm builds (where Go module cache + earlier layers can be reused) typically drop to ~5–10 minutes per image. Per-image scoping prevents cache collisions between the two matrix variants.

### 3. Fix `e2e-helm-install` image-tag override (the actual bug)

The chart's `values.yaml` ships with:

```yaml
controllerManager:
  container:
    image:
      tag: __PLACEHOLDER__
```

The placeholder is substituted only in the **tag-push** code path:

```yaml
- name: Package and push Helm chart
  if: github.ref_type == 'tag'
  run: |
    yq -i '.controllerManager.container.image.tag = strenv(chartVersion)' values.yaml
    helm package . ...
```

The mutation lives only in that job's working directory. On **main pushes**, the `e2e-helm-install` job re-checks-out the repo (pristine `__PLACEHOLDER__`), runs `helm install kagenti-operator ./charts/kagenti-operator` with no `--set`, and the rendered Deployment tries to pull `ghcr.io/.../kagenti-operator:__PLACEHOLDER__` — `ErrImagePull` → 120s rollout timeout.

Every recent main-push run of this workflow has been failing for this reason (PRs #366, #369, #370, #373). This means the e2e step has been providing **zero coverage** — it can't catch real regressions when it always fails for the same trivial reason.

Fix: pass `--set controllerManager.container.image.tag=latest`. The build-and-push job tags `:latest` on default-branch pushes, so the e2e installs whatever was just built.

## Bonus changes

- **Split release-only steps into a `release` job** (gated `if: github.ref_type == 'tag'`). The original sequence-of-steps shape doesn't survive matrix-ization since the helm-chart packaging and `gh release create` shouldn't fan out per image. The `release` job runs once, after both matrix jobs complete.

- **Idempotent `Create GitHub Release`**: treat `Release.tag_name already exists` (HTTP 422) as a benign no-op. This handles the case where a maintainer ran `gh release create` out of band before pushing the tag (e.g., for custom release notes).

## Test plan

This PR's branch is on a feature branch in the fork; pushing the PR will trigger the modified workflow on a `main` push (the merge commit). Expected outcomes:

- [ ] `build-and-push` matrix produces 2 parallel jobs (kagenti-operator, agentcard-signer)
- [ ] Each finishes in 5–22 minutes (cold cache the first time, warmer thereafter)
- [ ] `e2e-helm-install` runs and the Deployment becomes Ready (because `:latest` resolves)
- [ ] No `release` job (since this is a main push, not a tag push)

Once merged, the next maintainer-cut tag will exercise the `release` job (helm chart push + `gh release create`).

## Sequencing

This PR is independent of any other in-flight work. Can land at any time.

Assisted-By: Claude Code

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,8 +22,21 @@ env:
   CHARTS_PATH: ./charts
 
 jobs:
+  # Build the two images in parallel, one matrix instance per image.
+  # Each instance carries its own buildx cache scope so warm builds
+  # only re-pull what changed for that image.
   build-and-push:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        image_config:
+          - name: kagenti-operator
+            context: ./kagenti-operator
+            dockerfile: ./kagenti-operator/Dockerfile
+          - name: agentcard-signer
+            context: ./kagenti-operator
+            dockerfile: ./kagenti-operator/cmd/agentcard-signer/Dockerfile
     steps:
     - name: Checkout
       uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


pdettori (approved):
Clean, well-structured CI refactoring. Three solid improvements:

1. **Parallel matrix builds** — correct `fail-fast: false`, per-image cache scoping prevents collisions, `needs: build-and-push` gates the release job properly.
2. **buildx GHA cache** — per-image `scope` is the right approach for matrix builds sharing the same Actions cache namespace.
3. **E2e fix** — `--set controllerManager.container.image.tag=latest` correctly resolves the `__PLACEHOLDER__` issue on main-push runs.

The `gh release view` precheck is cleaner than stderr-grepping — good response to review feedback.

**Areas reviewed**: CI/GitHub Actions, Shell scripting, Security (action pinning)
**Commits**: 2 commits, all signed-off ✓
**CI status**: All 15 checks passing ✓

PR #380, opened 2026-05-27
Card Discovery Refinement Alignment [Spec + Impl]

rhuss:
> **[Review Guide](https://github.com/rhuss/kagenti-operator/blob/002-card-discovery-alignment/specs/002-card-discovery-alignment/REVIEWERS.md)** for full context: motivation, key decisions, and scope boundaries.

## Spec for review

PR #372 introduced card discovery on AgentRuntime. This follow-up aligns the API surface with Kubernetes conventions and the community refinement document before external consumers lock in the current field names and condition types.

### Why now

The card discovery API has no consumers yet. Field renames and condition type changes are free today but require migration tomorrow. This is the last window to get the API right before it becomes a compatibility constraint.

### Proposed changes (spec only, no code yet)

**Transport security visibility** (new `transportSecurity` field on CardStatus)

A card fetched over plain HTTP within the cluster network has no transport-level identity guarantee. Platform engineers currently cannot tell from the AgentRuntime status whether a card was fetched securely. Adding `transportSecurity` ("mTLS" or "plainHTTP") makes the security posture observable without digging into operator logs.

**Condition model cleanup** (rename `CardSynced` to `CardFetched`)

- `CardSynced` implies ongoing synchronization, which is misleading for an event-driven, non-polling architecture
- `CardSynced True CardSynced` stutters in `kubectl describe` output (condition type and reason are identical)
- Transport-aware reasons (`Fetched` vs `FetchedInsecure`) give immediate security posture visibility
- `WorkloadNotReady` split from `ServiceNotFound` maps each reason to one diagnostic action

**Field name accuracy**

- `cardId` renamed to `cardHash`: the field holds a SHA-256 content hash for change detection, not a unique identifier
- `fetchedAt` renamed to `lastCardFetchTime`: disambiguates "when the controller fetched it" from "when the agent created it"

**Protocol-aware port resolution**

- `kagenti.io/port` annotation override for multi-port Services where auto-detection picks the wrong port
- Port named `a2a` takes priority over generic `http` since we're fetching A2A protocol cards specifically

### Artifacts

| File | Purpose |
|------|---------|
| spec.md | 4 user stories, 9 functional requirements, 5 success criteria |
| plan.md | Technical context, constitution check, file map |
| tasks.md | 31 tasks across 6 phases |
| research.md | 6 research decisions |
| data-model.md | CardStatus field changes, condition model |
| REVIEWERS.md | Review guide |

Assisted-By: 🤖 Claude Code


specs/002-card-discovery-alignment/REVIEWERS.md
```diff
@@ -0,0 +1,83 @@
...
+
+- **Breaking changes**: All field renames and the condition type rename are CRD-breaking. This is acceptable because PR #372 just merged with no external consumers yet. If any downstream tooling has already started parsing `cardId` or `CardSynced`, this will break it.
+
+- **ConfigMap fetch path**: The `transportSecurity` value for the ConfigMap path (signed card from init-container) is `"configMap"`. This path is being deprecated but still exists during coexistence. Reviewers should assess whether this value is appropriate or whether `"signed"` better describes the semantics.
+
+- **Constitution update**: The project constitution (`.specify/memory/constitution.md`) references `CardSynced` from the original bug fix. It needs to be updated to `CardFetched` as part of this change (FR-009).
+
+## Open Questions
+
+1. Should `transportSecurity` values be documented as a formal enum in the CRD description, or left as free-form strings for extensibility (e.g., future `"ztunnel"` value)?
```
pdettori, line 69: **suggestion**: Strong recommendation to make this a validated enum (`+kubebuilder:validation:Enum=mTLS;plainHTTP;configMap`) rather than a free-form string.

Reasoning: this field's primary consumers are policy engines and compliance dashboards that will do exact-match comparisons. A free-form string invites silent divergence (e.g., `"mtls"` vs `"mTLS"` vs `"mutual_tls"`). Future values like `"ztunnel"` can be added via CRD schema evolution — that's a normal v1alpha1 change. The extensibility cost of an enum is near-zero at this API maturity level, while the correctness benefit is immediate.

specs/002-card-discovery-alignment/spec.md
```diff
@@ -0,0 +1,126 @@
...
+- **FR-002**: The system MUST rename the condition type from `CardSynced` to `CardFetched`.
+- **FR-003**: The `CardFetched` condition MUST use these reasons: `Fetched` (mTLS success), `FetchedInsecure` (plain HTTP success), `FetchSkipped` (pod template unchanged), `FetchFailed` (fetch error), `ServiceNotFound` (no matching Service), `WorkloadNotReady` (no Ready pods), `DiscoveryDisabled` (feature flag off).
+- **FR-004**: The system MUST rename the `cardId` field to `cardHash` in `CardStatus`.
+- **FR-005**: The system MUST rename the `fetchedAt` field to `lastCardFetchTime` in `CardStatus`.
+- **FR-006**: The system MUST resolve the A2A card endpoint port using this chain: `kagenti.io/port` annotation on the Service (highest priority), then port named `a2a`, then port named `http`, then first port.
+- **FR-007**: The system MUST check workload readiness before attempting service resolution. For Deployments and StatefulSets, check `readyReplicas > 0`. For Sandboxes (unstructured), skip the readiness check and proceed directly to service resolution. If the readiness check fails, set `CardFetched=False` with reason `WorkloadNotReady`.
```
pdettori, line 101: **question**: How should scale-to-zero workloads behave?

If a workload intentionally scales to zero (KEDA, Knative, or manual `replicas: 0` for maintenance), this check will fire `WorkloadNotReady` permanently. That's arguably correct (the card *can't* be fetched), but it could be noisy for operators who intentionally scale down.

Worth considering:
1. Is permanent `WorkloadNotReady` acceptable for intentional scale-to-zero? (I think yes — it accurately reflects that the card is stale)
2. Should the spec explicitly state this expected behavior so implementers don't add a special case?

Not blocking, just asking the spec to be explicit about this edge case.

specs/002-card-discovery-alignment/data-model.md
```diff
@@ -0,0 +1,80 @@
...
+| Constant | Old Value | New Value |
+|----------|-----------|-----------|
+| `ConditionTypeCardSynced` | `"CardSynced"` | Renamed to `ConditionTypeCardFetched` = `"CardFetched"` |
+
+### Condition Reasons
+
+| Old Reason | New Reason | Status | When |
+|------------|------------|--------|------|
+| `"CardSynced"` | `"Fetched"` | True | mTLS fetch succeeded |
+| (new) | `"FetchedInsecure"` | True | Plain HTTP fetch succeeded |
```
pdettori, line 67: **suggestion**: Consider whether `FetchedInsecure` with `Status: True` deserves a printer column for `transportSecurity`.

Right now FR-008 replaces the printer column with `CardFetched` (True/False). A platform engineer running `kubectl get agentruntime` will see `CardFetched=True` for *both* mTLS and plain HTTP fetches — the security distinction is invisible without `-o yaml`.

A second printer column showing transport security (or a composite like `CardFetched: True (plainHTTP)` in the reason) would make the security posture scannable at a glance. Not essential for v1, but worth noting as a potential follow-up or discussing whether it belongs in this spec's scope.

specs/002-card-discovery-alignment/research.md
```diff
@@ -0,0 +1,94 @@
...
+
+**Rationale**: Follows the project's existing annotation namespace (`kagenti.io/type`, `kagenti.io/framework`). Placed on the Service (not the Deployment) because the port is a Service-level concern.
+
+**Alternatives considered**:
+- `agent.kagenti.dev/port`: uses the CRD API group, which is for CRD fields not annotations
+- `protocol.kagenti.io/a2a-port`: too specific, couples annotation to one protocol
+
+## R-003: Workload readiness check placement
+
+**Decision**: Check `readyReplicas > 0` on the workload object before calling `resolveServiceForWorkload`.
```
pdettori, line 31: **nit**: The spec says "skip readiness check for Sandboxes" — it would help to add a one-liner explaining *why* Sandboxes are exempt (e.g., "Sandboxes are unstructured resources without a standard readyReplicas field" or "Sandbox lifecycle is managed differently"). Makes the decision self-documenting for future readers.

pdettori (approved):
## Review Summary

This is a well-structured spec that makes the right architectural call at the right time. The API surface improvements are genuine — better naming, security observability, clearer diagnostics — and the cost is minimal given the absence of external consumers.

The synthesis approach is sound: cherry-picking transport security awareness and workload readiness from the refinement doc while preserving PR #372's event-driven model and `FetchSkipped` semantics. The timing argument for breaking changes is valid.

A few design questions below (none blocking), mostly around ensuring the spec is explicit on edge cases that will matter during implementation.

**Areas reviewed**: Data model, condition semantics, port resolution, workload readiness, CRD UX
**Commits**: 3, all signed-off
**CI**: All passing (E2E pending)

specs/002-card-discovery-alignment/REVIEWERS.md
rhuss: Agreed. Changed FR-001 to require `+kubebuilder:validation:Enum=mTLS;plainHTTP;configMap`. At v1alpha1, adding values is a normal schema evolution, so the extensibility cost is near-zero.

specs/002-card-discovery-alignment/spec.md
rhuss: Good catch. Added as an explicit edge case: `WorkloadNotReady` is correct and intentional for scale-to-zero. The card genuinely can't be fetched, and the condition accurately reflects stale data. No special-casing needed.

specs/002-card-discovery-alignment/data-model.md
rhuss: Good idea. A `TransportSecurity` printer column would make security posture scannable via `kubectl get` without `-o yaml`. Keeping it out of this spec's scope since it's purely additive and can be a quick follow-up. Noted for the implementation.

specs/002-card-discovery-alignment/research.md
rhuss: Added: "Sandboxes are unstructured resources without a standard `readyReplicas` field; their lifecycle is managed by the sandbox controller, not by Deployment/StatefulSet rollout semantics."

specs/002-card-discovery-alignment/spec.md
rhuss: But you are right, this is a tricky thing, as a scaled-down e.g Knative service is not directly ready but it still _active_ it the sense that it will process incoming requests (but requires a cold start). So, when we periodically fetch the agent card, we will keep up the Knative service alive anyway (but I guess autoscaling from zero is not enabled for infra structure ports, otherwise every Prometheus scrape would wake up a sleep service unnecessarily). I've added the edge case in more detail to the spec.

kagenti-operator/config/crd/bases/agent.kagenti.dev_agentruntimes.yaml
```diff
@@ -445,6 +469,14 @@ spec:
                       Indicates if the agent supports providing an extended agent card when
                       authenticated.
                     type: boolean
+                  transportSecurity:
+                    description: TransportSecurity indicates the transport layer used
+                      for the card fetch.
+                    enum:
+                    - mTLS
+                    - plainHTTP
+                    - configMap
```
rhuss: Not sure about this one. What is this `configMap` mode actually ? (i will check)

kagenti-operator/internal/controller/agentruntime_controller.go
```diff
@@ -825,12 +881,12 @@ func (r *AgentRuntimeReconciler) fetchCard(
...
 			if err != nil {
-				return nil, nil, fmt.Errorf("authenticated fetch failed for %s: %w", ref.Name, err)
+				return nil, nil, "", fmt.Errorf("authenticated fetch failed for %s: %w", ref.Name, err)
 			}
 			if fetchResult.CardData == nil {
-				return nil, nil, fmt.Errorf("authenticated fetch returned nil card data for %s", ref.Name)
+				return nil, nil, "", fmt.Errorf("authenticated fetch returned nil card data for %s", ref.Name)
 			}
-			return fetchResult.CardData, fetchResult, nil
+			return fetchResult.CardData, fetchResult, "mTLS", nil
```
rhuss: Shouldn't this be also used via an enum that we have defined before ?

kagenti-operator/internal/controller/agentruntime_controller.go
```diff
@@ -841,15 +897,18 @@ func (r *AgentRuntimeReconciler) fetchCard(
...
 	cardData, err := r.AgentFetcher.Fetch(ctx, protocol, serviceURL, ref.Name, rt.Namespace)
 	if err != nil {
-		return nil, nil, fmt.Errorf("fetch failed for %s: %w", ref.Name, err)
+		return nil, nil, "", fmt.Errorf("fetch failed for %s: %w", ref.Name, err)
+	}
+	if cardData == nil {
+		return nil, nil, "", fmt.Errorf("fetch returned nil card data for %s", ref.Name)
 	}
-	return cardData, nil, nil
+	return cardData, nil, "plainHTTP", nil
```
rhuss: Via enum ?

kagenti-operator/internal/controller/agentruntime_controller.go
```diff
@@ -726,39 +769,46 @@ func (r *AgentRuntimeReconciler) fetchAndUpdateCard(ctx context.Context, rt *age
...
+	newCardHash := computeCardContentHash(cardData)
 
 	cardStatus := &agentv1alpha1.CardStatus{
-		AgentCardData: *cardData,
-		CardID:        newCardID,
-		Protocol:      protocol,
+		AgentCardData:     *cardData,
+		CardHash:          newCardHash,
+		Protocol:          protocol,
+		TransportSecurity: transportSecurity,
```
rhuss, line 804: what's the difference between "Protocol" and "TransportSecurity" ?

kagenti-operator/config/crd/bases/agent.kagenti.dev_agentruntimes.yaml
rhuss: Removed. The `configMap` value was intended for the ConfigMap fetch path (init-container signing), but that path is managed by the AgentCard controller, not the AgentRuntime controller. Since this code never sets it, keeping it in the enum would be misleading. It can be added back if/when the ConfigMap path is migrated to AgentRuntime.

Enum is now `mtls` and `http` only (commit 2ba4372).

kagenti-operator/internal/controller/agentruntime_controller.go
rhuss: Fixed. Introduced a typed `TransportSecurity` enum in `agentruntime_types.go` with constants `TransportSecurityMTLS` (`"mtls"`) and `TransportSecurityHTTP` (`"http"`). All string literals in the controller now use these constants. Also renamed the values to lowercase to follow Kubernetes conventions (commit 2ba4372).

kagenti-operator/internal/controller/agentruntime_controller.go
rhuss: Fixed in the same commit. Now uses `agentv1alpha1.TransportSecurityHTTP` (commit 2ba4372).

kagenti-operator/internal/controller/agentruntime_controller.go
rhuss: They are orthogonal concepts:

- `protocol` = the application-level agent discovery protocol (`"a2a"`), identifying which card endpoint convention was used
- `transportSecurity` = the transport layer authentication used for the fetch (`"mtls"` or `"http"`)

An A2A card can be fetched over either transport. `protocol` tells you *what* was fetched, `transportSecurity` tells you *how securely* it was fetched.

PR #386, opened 2026-05-29
Feat: detect and warn when OVN routingViaHost is not enabled

Bobbins228:
## Summary

Adds a startup-time network configuration check that detects when OVN-Kubernetes `routingViaHost` is not enabled on OpenShift clusters. Without this setting, Istio ambient mode's ztunnel cannot intercept pod-to-pod traffic and mTLS/authorization policies are silently bypassed. The operator logs a warning at startup — no per-CR conditions, no per-reconcile overhead.

## Changes

- **`network_check.go`** — `NetworkOperatorCRDExists` (CRD discovery at startup) and `CheckOVNNetworkConfig` (reads `network.operator.openshift.io/cluster`, returns a warning string if misconfigured)
- **`cmd/main.go`** — runs the check once at startup after CRD discovery, logs warning if misconfigured
- **RBAC** — `get` on `operator.openshift.io/networks` in `config/rbac/role.yaml` and Helm chart
- **`network_check_test.go`** — 5 unit tests covering misconfigured OVN, correct config, non-OVN clusters, and missing resource

### DIFF
--- a/kagenti-operator/cmd/main.go
+++ b/kagenti-operator/cmd/main.go
@@ -21,6 +21,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"flag"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -480,6 +481,14 @@ func main() {
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "AgentCardSync")
 		os.Exit(1)
+	}
+
+	if controller.NetworkOperatorCRDExists(mgr.GetConfig()) {
+		if warning := controller.CheckOVNNetworkConfig(context.Background(), mgr.GetAPIReader()); warning != "" {
+			setupLog.Error(fmt.Errorf("OVN network misconfiguration"), warning)
+		} else {
+			setupLog.Info("OVN-Kubernetes routingViaHost is correctly configured")
+		}
 	}
 
 	artReconciler := &controller.AgentRuntimeReconciler{
--- a/kagenti-operator/internal/controller/agentruntime_controller.go
+++ b/kagenti-operator/internal/controller/agentruntime_controller.go
@@ -124,6 +124,7 @@ func (r *AgentRuntimeReconciler) getFeatureGates() *webhookconfig.FeatureGates {
 // +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
 // +kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch
 // +kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
+// +kubebuilder:rbac:groups=operator.openshift.io,resources=networks,verbs=get
 
 func (r *AgentRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
--- a/kagenti-operator/internal/controller/network_check.go
+++ b/kagenti-operator/internal/controller/network_check.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"fmt"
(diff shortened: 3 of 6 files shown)


kagenti-operator/internal/controller/agentruntime_controller.go
```diff
@@ -173,6 +176,9 @@ func (r *AgentRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 	}
 
+	// 4.2. Check OVN network configuration (OpenShift only).
+	r.checkNetworkConfig(ctx, rt)
```
ChristianZaccaria: We should reconsider the approach here. For context, the network check is a cluster-level fact and not per-workload concern. It doesn't vary per agent or runtime, hence the agenruntime_controller shouldn't be responsible for performing this check.

Moreover, with the changes in agentruntime_controller, this would mean per N agents, you get N API calls on every reconcile which can be a lot.

Simpler approach: Do the check once at startup + log warning if necessary in main.go:

Startup check + log warning. Run CheckOVNNetworkConfig once in main.go right after NetworkOperatorCRDExists. Log a clear warning. Expected around <50 lines of code total, zero per-reconcile overhead, no per-CR conditions to maintain. Operators see it in the pod logs.

kagenti-operator/internal/controller/network_check.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+var networkOperatorGVK = schema.GroupVersionKind{
+	Group:   "operator.openshift.io",
+	Version: "v1",
+	Kind:    "Network",
+}
+
+// NetworkCheckResult represents the outcome of checking OVN network configuration.
+type NetworkCheckResult struct {
```
ChristianZaccaria: We can remove the whole struct:
Since we probably won't propagate the conditions or messages to the AgentRuntime CR, we can remove them entirely, and just return a simple string to be logged by the operator.

ChristianZaccaria (commented):
Thanks for tackling this, we may need to reconsider the approach, please see comments below for more info.

ChristianZaccaria (changes requested):

Bobbins228:
Addressed both review comments:

1. **Moved check to startup** — `CheckOVNNetworkConfig` now runs once in `main.go` right after `NetworkOperatorCRDExists`. No per-reconcile overhead, no per-CR conditions to maintain. The reconciler is untouched.

2. **Removed `NetworkCheckResult` struct** — `CheckOVNNetworkConfig` returns a plain warning string (empty = no issue). ~350 lines removed from the PR.

RBAC narrowed to just `get` (single read at startup, no list/watch needed).

ChristianZaccaria (approved):
/lgtm !!

ChristianZaccaria (commented):
Please rebase when you have a chance, then should be good to merge!

kagenti-operator/cmd/main.go
```diff
@@ -427,6 +427,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	if controller.NetworkOperatorCRDExists(mgr.GetConfig()) {
+		if warning := controller.CheckOVNNetworkConfig(context.Background(), mgr.GetAPIReader()); warning != "" {
+			setupLog.Info("OVN network configuration warning", "detail", warning)
```
cwiklik: **nit (optional):** This misconfiguration silently disables Istio ambient mTLS/authz — a security-relevant condition — but it's surfaced only at `Info` level (`setupLog.Info("OVN network configuration warning", …)`), where it can be lost in startup noise. logr has no `Warn`, so this is idiomatic, but consider making it stand out (a distinct, unmissable log line, or a Kubernetes Event if it ever warrants more). Consistent with the deliberately lightweight startup-only design you described, so entirely optional.

cwiklik (approved):
Clean, low-risk, well-tested startup check: detects OVN-Kubernetes without `routingViaHost` (which silently breaks ambient mTLS) and logs an actionable warning with the exact remediation patch. Correctly OCP-gated via `NetworkOperatorCRDExists`, so it no-ops on non-OpenShift clusters. The read-only `operator.openshift.io/networks get` permission is added in all three places — the kubebuilder marker, `config/rbac/role.yaml`, and the chart `role.yaml`, kept in sync (the hygiene that avoids RBAC drift). Five unit tests cover OVN-unset/false/true, non-OVN, and missing-resource paths.

**CI — E2E is red, but the failure is unrelated:** `Manager … should ensure the metrics endpoint is serving metrics` timed out (300s) at `test/e2e/e2e_test.go:249`; 31/32 specs passed. The OVN code can't affect it — it no-ops on the Kind E2E cluster, and the only other changes are a gofmt whitespace fix and an RBAC marker. Recommend re-running E2E so branch protection goes green before merge; not a code blocker.

**Security:** read-only `get`; no secrets; the check hardens a silent mTLS-bypass condition.

**Commits:** signed-off. Nit: `feat:` subject lacks the emoji prefix Kagenti's convention favors (style only).

pdettori:
@Bobbins228 need rebase

PR #332, opened 2026-05-06
Add OCI skill image mounting to AgentRuntime

cooktheryan:
## Summary

- Adds a `skills` field to `AgentRuntimeSpec` for declaring OCI skill images to mount into agent pods as Kubernetes [ImageVolumes](https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes/)
- Gated behind a `skillImageVolumes` feature gate (default off), requires Kubernetes 1.31+
- Uses the [skillimage](https://github.com/redhat-et/skillimage) OCI format: `FROM scratch` images with `skill.yaml` + `SKILL.md`
- Each skill specifies a `mountPath`, making the feature framework-agnostic (Claude, Cursor, custom agents, etc.)

### Example

```yaml
apiVersion: agent.kagenti.dev/v1alpha1
kind: AgentRuntime
metadata:
  name: resume-agent-runtime
spec:
  type: agent
  targetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: resume-agent
  skills:
    - name: resume-reviewer
      image: ghcr.io/redhat-et/skillimage/resume-reviewer:v1.0.0
      mountPath: /agent/skills/resume-reviewer
    - name: blog-writer
      image: ghcr.io/redhat-et/skillimage/blog-writer:latest
      mountPath: /app/.claude/skills/blog-writer
      pullPolicy: Always
```

### Changes

| Area | Files | What |
|------|-------|------|
| CRD types | `api/v1alpha1/agentruntime_types.go` | `SkillImageRef`, `SkillPullPolicy`, `Skills` field |
| Feature gate | `internal/webhook/config/feature_gates.go` | `SkillImageVolumes` (default false) |
| Controller | `internal/controller/agentruntime_controller.go`, `agentruntime_skills.go` | Reconcile ImageVolumes on Deployment/StatefulSet, cleanup on deletion |
| Config hash | `internal/controller/agentruntime_config.go` | Skills in hash → rolling updates on change |
| Webhook | `internal/webhook/v1alpha1/agentruntime_webhook.go` | Validate duplicate names, reserved volume collisions |
| Wiring | `cmd/main.go` | Pass feature gate loader to reconciler |
| Docs | `docs/api-reference.md`, `docs/architecture.md` | SkillImageRef reference, conditions, examples |
| Samples | `config/samples/agent_v1alpha1_agentruntime_skills.yaml`, updated `_full.yaml` | New and updated sample manifests |
[...]

PR #388, opened 2026-05-29
feat: replace spec.skills injection with passive skill discovery

cooktheryan:
## Summary

- Remove `spec.skills` from the AgentRuntime CRD — the operator no longer mutates target Deployments to inject OCI skill ImageVolumes, eliminating GitOps drift with ArgoCD/Flux
- Skills are now declared directly in the Deployment manifest (OCI ImageVolumes or ConfigMap volumes) and discovered by the operator through the `kagenti.io/skills` annotation
- New `skillDiscovery` feature gate (default: false) controls whether the operator reads the annotation and populates `status.linkedSkills`
- The agent discovers mounted skills via `SKILL_FOLDERS` and reports them in its A2A card — the operator reads the card and the annotation, surfacing both in status

### Context

PR #332 added `spec.skills` to AgentRuntime, which instructed the operator to inject OCI ImageVolumes into the target Deployment via `r.Update()`. This caused GitOps drift (ArgoCD/Flux fight with the operator over the Deployment spec) and was architecturally overreaching — the operator was modifying a workload it doesn't own.

This PR changes the operator's role from **injector** to **observer**:
- **Deployment** owns its skill volumes (user-managed, in Git)
- **Agent** discovers mounted skills at runtime via `SKILL_FOLDERS` and advertises them in its A2A card
- **AgentRuntime** reads the `kagenti.io/skills` annotation and the card, reports both in status

This aligns with the ConfigMap-based skill path in kagenti/kagenti#1440, where the kagenti backend sets the same annotation. Both skill delivery mechanisms (ConfigMap and OCI ImageVolume) are now visible through a single annotation and the agent's card.

## Test plan

- [x] `go build ./...` — compiles cleanly
- [x] `go vet ./...` — no issues
- [x] `golangci-lint run ./...` — no new issues
- [x] Kind cluster validation: Deployment with both OCI ImageVolume and ConfigMap skills, AgentRuntime discovers both via annotation, agent card reflects all mounted skills
- [x] Feature gate validation: `skillDiscovery: true` populates `status.linkedSkills`, `skillDiscovery: false` clears it
- [x] Operator does NOT mutate the Deployment PodSpec
- [x] E2E tests (skill injection block removed, needs new skill discovery e2e)

Assisted-By: Claude Code


cooktheryan:
@pavelanni @kevincogan @pdettori @eranra I wanted to re-roll the way skills were defined. I started the process of blogging about the skill work and I quickly realize we would make for a really painful scenario with GitOps/ArgoCD with modifying the volume mounts of the deployment. I wanted to try to pull in the skills dynamically while also referencing what was brought in with https://github.com/kagenti/kagenti/pull/1440 I am open to any feedback, conversations, and etc. @kevincogan I also want to ensure this does not cause issues with any security related ideas or concepts you had.

One of the big things I am thinking about with skills is the fact that a user may start with 0 skills attached. Then add a skill or two. Lifecycle it for a while then discover possibly the skills is not longer required. All of this could and should occur without the need to rewrite or rebuild the Agent unless it is necesary.


```
apiVersion: agent.kagenti.dev/v1alpha1
kind: AgentRuntime
metadata:
  annotations:
    agent.kagenti.dev/last-card-fetch-hash: "5"
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"agent.kagenti.dev/v1alpha1","kind":"AgentRuntime","metadata":{"annotations":{},"labels":{"app.kubernetes.io/name":"a2a-currency-converter"},"name":"a2a-currency-converter","namespace":"default"},"spec":{"targetRef":{"apiVersion":"apps/v1","kind":"Deployment","name":"a2a-currency-converter"},"type":"agent"}}
  creationTimestamp: "2026-05-29T17:55:10Z"
  finalizers:
  - kagenti.io/cleanup
  generation: 1
  labels:
    app.kubernetes.io/name: a2a-currency-converter
  name: a2a-currency-converter
  namespace: default
  resourceVersion: "5688"
  uid: fccb4363-7873-4a05-992e-8dca085404c4
spec:
  targetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: a2a-currency-converter
  type: agent
status:
  card:
    capabilities:
      pushNotifications: false
      streaming: false
    cardId: 294e46254525e6aa0e0a10969044437cd7fabfcd9b97ff76385a1f534588d5d3
    description: Converts between currencies using real-time exchange rates
    fetchedAt: "2026-05-29T17:55:10Z"
    name: Currency Converter
    protocol: a2a
    skills:
    - description: Convert an amount from one currency to another
      examples:
      - Convert 100 USD to EUR
      - How much is 50 GBP in JPY?
      id: convert-currency
      name: Currency Conversion
    - description: Review Kubernetes manifests for OpenShift readiness
      id: openshift-review
      name: openshift-review
    - description: Summarize long text into concise structured summaries
      id: summarizer
      name: summarizer
    url: http://a2a-currency-converter.default.svc.cluster.local:8080/
    version: 1.0.0
  conditions:
  - lastTransitionTime: "2026-05-29T17:55:10Z"
    message: Deployment a2a-currency-converter resolved
    observedGeneration: 1
    reason: TargetFound
    status: "True"
    type: TargetResolved
  - lastTransitionTime: "2026-05-29T17:55:10Z"
    message: Configuration resolved successfully
    observedGeneration: 1
    reason: ConfigResolved
    status: "True"
    type: ConfigResolved
  - lastTransitionTime: "2026-05-29T17:55:10Z"
    message: Pod template unchanged; existing card data still valid
    observedGeneration: 1
    reason: FetchSkipped
    status: "True"
    type: CardSynced
  - lastTransitionTime: "2026-05-29T17:55:10Z"
    message: 2 linked skill(s) discovered from workload annotation
    observedGeneration: 1
    reason: SkillsFound
    status: "True"
    type: SkillsDiscovered
  - lastTransitionTime: "2026-05-29T17:55:10Z"
    message: Workload a2a-currency-converter configured with config-hash 57de3ab75a7e
    observedGeneration: 1
    reason: Configured
    status: "True"
    type: Ready
  configuredPods: 2
  linkedSkills:
  - openshift-review
  - summarizer
  phase: Active
  ```

**Comment by pavelanni** [comment]
Adding skills dynamically is possible (you can add a tool `skill-puller` to AgentRuntime) but that breaks the idea of having read-only skills to avoid malicious agents/skills to edit existing skills. When we _mount_ them upfront read-only we are safe. When the agent pulls them during its activity -- we have to make sure that 1) it verifies the skill (based on its signature or SHA) and 2) it locks the pulled skill so it can't change it after that. The second part is tricky. Let me think about it and investigate.
Alternatively, we can define different modes in which AgentRuntime operates. One is flexible/development/unsafe mode, and another is locked-down, production mode (where all skills are mounted read-only upfront). It is also good from the positioning perspective when you give the user a choice and the user understands the risks.

**Comment by cooktheryan** [comment]
looking forward to what find @pavelanni I think one of the pieces I really overlooked was around how people manage their running applications. We definitely need to present and provide our solution with as much security around it as possible

**Comment by pavelanni** [comment]
After reviewing several options I think the right solution for dynamic skill discovery and deployment will be a `skill-puller` sidecar that can search, verify, and pull skills into the shared volume that is mounted read-write for this sidecar and read-only for the main agent. The sidecar itself will be written in Go so it won't add much to the total pod's resource consumption. It will be possible to configure the agent Deployment with this sidecar depending whether we need dynamic agents or not.
I'll create a prototype next week.

**Comment by kevincogan** [comment]
> @kevincogan I also want to ensure this does not cause issues with any security related ideas or concepts you had.

@cooktheryan No conflicts with the mTLS/identity work. The observer model is cleaner since the security controls operate on identity and the card, not on how skills are delivered.

**Comment by kevincogan** [comment]
> After reviewing several options I think the right solution for dynamic skill discovery and deployment will be a `skill-puller` sidecar that can search, verify, and pull skills into the shared volume that is mounted read-write for this sidecar and read-only for the main agent. The sidecar itself will be written in Go so it won't add much to the total pod's resource consumption. It will be possible to configure the agent Deployment with this sidecar depending whether we need dynamic agents or not. I'll create a prototype next week.

@pavelanni what use cases are you targeting with the sidecar? Are there agents today that can't tolerate rolling restarts on skill changes, or is this more forward-looking?

**Comment by pavelanni** [comment]
@kevincogan 
> what use cases are you targeting with the sidecar? Are there agents today that can't tolerate rolling restarts on skill changes, or is this more forward-looking?

The use case is when the user wants to add skills dynically but without changing the AgentRuntime config and while keeping the skills volume read-only. The new skill is added in two steps: 1) it is discovered and pulled into the volume by the sidecar; 2) the agent re-reads the skills directory and adds the new skill to its list of skills.
If the agent can't re-read the skills, then yes, it will require a restart.

**Comment by cooktheryan** [comment]
@pavelanni could we cover the dynamic reloading in a follow up PR. From the OCI perspective if we modify the deployment then its going to cause the harness to reload the container by default. I aded a test case for that to ensure agentruntime is updated when a reload occurs due to an OCI skill being removed

File: kagenti-operator/internal/controller/agentruntime_controller.go [hunk 1]
```diff
@@ -510,6 +473,43 @@ func (r *AgentRuntimeReconciler) countConfiguredPods(ctx context.Context, rt *ag
...
+
+	raw, ok := annotations[AnnotationSkills]
+	if !ok || raw == "" {
+		return nil
+	}
+
+	var skills []string
+	if err := json.Unmarshal([]byte(raw), &skills); err != nil {
+		logger.V(1).Info("Failed to parse kagenti.io/skills annotation", "error", err, "raw", raw)
+		return nil
```
pdettori, line 512: **suggestion**: The V(1) log on malformed annotation JSON is likely invisible in most deployments. Consider also emitting a Kubernetes Event via `r.Recorder.Event(rt, corev1.EventTypeWarning, "SkillAnnotationParseError", ...)` so users get actionable feedback when their `kagenti.io/skills` annotation isn't valid JSON — similar to how other error paths in this controller use the recorder.

kagenti-operator/test/e2e/e2e_test.go
```diff
@@ -1941,7 +1941,7 @@ rules:
 	})
 })
 
-var _ = Describe("Skill Image Volumes E2E", Ordered, func() {
+var _ = Describe("Skill Discovery E2E", Ordered, func() {
```
pdettori, line 1944: **suggestion**: Consider adding a test case for a malformed `kagenti.io/skills` annotation (e.g., `"not-valid-json"`) to verify the reconcile loop degrades gracefully (empty linkedSkills, no crash). Would be a cheap addition alongside the existing gate-on/gate-off coverage.

pdettori (approved):
Clean architectural improvement: moves from operator-mutates-workloads (GitOps drift) to passive annotation-based skill discovery. Feature gate properly defaults to false, CI all green, E2E covers both gate states. Two suggestions below (non-blocking).

cooktheryan:
@kevincogan @pavelanni @Ladas can I get an additional review here. I patched up @pdettori concerns

kevincogan (approved):
Traced through the reconcile paths, feature gate lifecycle, and E2E coverage. LGTM

Ladas (approved):
Looks good, seems like the concerns like the dynamic reload can be addressed in followup with proper tests.

pavelanni:
looks good to me

PR #395, opened 2026-06-02
fix(client-registration): attach audience scope to agent's own client

### LINKED ISSUE #394: bug: ClientRegistrationReconciler creates audience-mapper without included.client.audience; scope not attached to existing clients
## Summary

The operator's `EnsureAudienceScope` (in `kagenti-operator/internal/keycloak/audience.go`, called from `ClientRegistrationReconciler`) creates the audience client-scope and registers an `oidc-audience-mapper`, but the mapper is created with **`included.client.audience` unset**. As a result, tokens issued via `client_credentials` for the agent's client carry `aud=<realm-URL>` instead of `aud=<SPIFFE-ID>`, and AuthBridge's `jwt-validation` plugin rejects them with `"aud" not satisfied` (its default `audience_file=/shared/client-id.txt` expects the SPIFFE ID).

There is also a **secondary bug**: even if the mapper were correct, the scope isn't attached to clients that were registered **before** it was added to the realm's default-default-client-scopes — so the basic weather-agent demo's CLI test (`demo-ui.md` step 6d) fails until both are fixed.

## Affected version

`kagenti/kagenti-operator` `v0.2.0-rc.5` (the version deployed by the current Kagenti installer in our default Kind setup). Likely affects all v0.2.x releases that include the operator-managed client-registration controller.

## Reproducer (basic weather-agent demo)

1. Deploy a fresh Kagenti install to Kind via `./.github/scripts/local-setup/kind-full-test.sh`.
2. Follow [`authbridge/demos/weather-agent/demo-ui.md`](https://github.com/kagenti/kagenti-extensions/blob/main/authbridge/demos/weather-agent/demo-ui.md) to import the weather-service agent.
3. Run step 6d (CLI end-to-end test). The agent rejects the request with:
   ```
   {"error":"auth.unauthorized","message":"token validation failed","plugin":"jwt-validation"}
   ```
   AuthBridge logs:
   ```
   "JWT validation failed" error="validating JWT: \"aud\" not satisfied"
   ```

## Smoking-gun evidence (live cluster, kagenti-operator v0.2.0-rc.5)

**(1) The audience scope was created — that part works:**
```
$ curl .../admin/realms/kagenti/default-default-client-scopes | jq -r '.[].name'
openid
email
profile
roles
web-origins
kagenti-platform-audience
agent-team1-weather-service-aud   <-- created by operator
```

**(2) But the scope's audience-mapper has `included.client.audience` unset:**
```json
{
  "name": "agent-team1-weather-service-aud",
  "protocolMapper": "oidc-audience-mapper",
  "included_audience": null   <-- should be "spiffe://localtest.me/ns/team1/sa/weather-service"
}
```

**(3) And the scope is NOT attached to the agent's client:**
```
$ curl .../admin/realms/kagenti/clients/<agent-uuid>/default-client-scopes | jq -r '.[].name'
web-origins
openid
profile
kagenti-platform-audience
roles
email
                                       <-- agent-team1-weather-service-aud is missing
```

The client was registered before the realm-level default-default scope was added. Keycloak only auto-attaches default-default scopes to **new** clients, so existing clients miss out.

**(4) Tokens therefore lack the SPIFFE audience:**
```
$ # client_credentials token issued to the agent's client
$ ... | jq '.aud'
"http://keycloak.localtest.me:8080/realms/kagenti"
```

(Expected: an array containing `spiffe://localtest.me/ns/team1/sa/weather-service`.)

## Manual fix that confirms the diagnosis

Setting `config["included.client.audience"]` on the existing mapper and attaching the scope to the agent's client makes step 6d pass end-to-end. Verified on the live cluster:

```json
{
  "iss": "http://keycloak.localtest.me:8080/realms/kagenti",
  "aud": [
    "spiffe://localtest.me/ns/team1/sa/weather-service",
    "http://keycloak.localtest.me:8080/realms/kagenti"
  ],
  "azp": "spiffe://localtest.me/ns/team1/sa/weather-service"
}
```

Agent accepts the token and returns a normal A2A response.

## Suggested fix

Two things in `EnsureAudienceScope` (`kagenti-operator/internal/keycloak/audience.go`):

1. **Populate the mapper config on creation.** When `getOrCreateAudienceClientScope` creates the `oidc-audience-mapper`, ensure the request payload includes:
   ```json
   {
     "name": "<scope-name>",
     "protocol": "openid-connect",
     "protocolMapper": "oidc-audience-mapper",
     "config": {
       "included.client.audience": "<SPIFFE-ID>",
       "id.token.claim": "false",
       "access.token.claim": "true"
     }
   }
   ```
   And, idempotency: when the mapper already exists with empty/null `included.client.audience`, PUT-update it.

2. **Attach the scope to the client at registration time, not just at realm-default level.** The `ClientRegistrationReconciler` should explicitly PUT `clients/<uuid>/default-client-scopes/<scope-id>` after creating the client, so existing-client / scope-added-after-client races don't leave clients without the audience scope.

## Workaround

Until fixed, demos and CI tests can apply the manual patch (see [my reproducer script](#) — happy to attach if helpful). Two API calls per agent client:

```
PUT /admin/realms/{realm}/client-scopes/{scope-id}/protocol-mappers/models/{mapper-id}
   body: {..., config: {included.client.audience: <SPIFFE-ID>, access.token.claim: "true", ...}}

PUT /admin/realms/{realm}/clients/{client-uuid}/default-client-scopes/{scope-id}
```

## Related

- kagenti-extensions PR #439 — surfaced this while reviewing the basic weather-agent demo. The PR will keep its scope (doc fixes); this issue tracks the operator-side fix.
- kagenti/kagenti issue #1775 — sidecar injection mode docs (separate concern).

mrsabath:
## Summary

`EnsureAudienceScope` creates the audience client-scope and registers it as a realm `default-default-client-scope`, expecting Keycloak to auto-attach it to every client. **But Keycloak only auto-attaches default-default scopes to *newly-created* clients.** The agent's client is registered just before `EnsureAudienceScope` is called, so by the time the realm-default is set up, the agent client has already been minted without it.

As a result, `client_credentials` tokens for the agent carry `aud=<realm-URL>` instead of `aud=<SPIFFE-ID>`, and AuthBridge's `jwt-validation` plugin rejects them with `"aud" not satisfied`.

This affects the basic weather-agent demo (kagenti-extensions [`demo-ui.md` step 6d](https://github.com/kagenti/kagenti-extensions/blob/main/authbridge/demos/weather-agent/demo-ui.md)) end-to-end test, and any other workflow that issues `client_credentials` tokens directly to operator-managed agent clients.

## Fix

- Add `AgentClientUUID` to `AudienceParams`. When set, `EnsureAudienceScope` explicitly `PUT`s `clients/<uuid>/default-client-scopes/<scope-id>` for the agent's own client (alongside the existing platform-client and realm-default attachments).
- `ClientRegistrationReconciler` captures the internal UUID returned by `RegisterOrFetchClientWithToken` and passes it through.
- Two new unit tests in `audience_test.go`:
  - `TestEnsureAudienceScope_AttachesToAgentClient` — asserts the agent attachment happens when the UUID is set.
  - `TestEnsureAudienceScope_NoAgentAttachWhenUUIDEmpty` — asserts prior behavior is preserved (no spurious `PUT`s) when the UUID is empty.

## Verification

End-to-end on a Kind cluster running `kagenti-operator` v0.2.0-rc.5:

**Before** the fix (manually applied — bare client, no scope attached):
```json
{
  "iss": "http://keycloak.localtest.me:8080/realms/kagenti",
  "aud": "http://keycloak.localtest.me:8080/realms/kagenti"
}
```
Agent: `{"error":"auth.unauthorized","message":"token validation failed","plugin":"jwt-validation"}`
AuthBridge log: `"JWT validation failed" error="validating JWT: \"aud\" not satisfied"`

**After** the fix (proven by manually applying the equivalent two API calls in-cluster as a proof-of-concept):
```json
{
  "iss": "http://keycloak.localtest.me:8080/realms/kagenti",
  "aud": [
    "spiffe://localtest.me/ns/team1/sa/weather-service",
    "http://keycloak.localtest.me:8080/realms/kagenti"
  ],
  "azp": "spiffe://localtest.me/ns/team1/sa/weather-service"
}
```
Agent accepts the token and returns a normal A2A response.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/keycloak/...` (existing + 2 new tests pass)
- [x] `go test $(go list ./... | grep -v /e2e)` with envtest binaries (all green)
- [x] Manual in-cluster verification via the equivalent REST calls (token now contains the SPIFFE audience; agent accepts the request)

Fixes #394

### DIFF
--- a/kagenti-operator/internal/keycloak/audience.go
+++ b/kagenti-operator/internal/keycloak/audience.go
@@ -16,6 +16,8 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // AudienceParams configures audience client-scope management (mirrors AuthBridge client_registration.py).
@@ -82,7 +84,10 @@ func (a *Admin) EnsureAudienceScope(ctx context.Context, token string, p Audienc
 	// not retroactively apply realm default-default-client-scopes to clients that already exist
 	// when the scope is added, and the agent's client is registered just before this call.
 	if p.AgentClientUUID != "" {
-		_ = a.putClientDefaultClientScope(ctx, token, p.Realm, p.AgentClientUUID, scopeID)
+		if err := a.putClientDefaultClientScope(ctx, token, p.Realm, p.AgentClientUUID, scopeID); err != nil {
+			log.FromContext(ctx).V(1).Info("agent client scope attach failed",
+				"clientUUID", p.AgentClientUUID, "scope", scopeName, "err", err.Error())
+		}
 	}
 	for _, plat := range p.PlatformClientIDs {
 		plat = strings.TrimSpace(plat)
--- a/kagenti-operator/internal/keycloak/audience.go
+++ b/kagenti-operator/internal/keycloak/audience.go
@@ -25,6 +25,13 @@ type AudienceParams struct {
 	AudienceClientID     string   // OAuth clientId / SPIFFE ID used as custom audience in the mapper
 	PlatformClientIDs    []string // Keycloak clientId strings (e.g. UI client), not internal UUIDs
 	AudienceScopeEnabled bool     // when false, EnsureAudienceScope is a no-op
+
+	// AgentClientUUID is the internal Keycloak UUID of the agent's own client. When non-empty,
+	// the audience scope is explicitly attached to this client as a default-client-scope, so
+	// tokens issued via client_credentials carry the audience claim. Without this, agents
+	// registered before the realm-level default-default scope existed won't have it attached
+	// (Keycloak only auto-attaches default-defaults to newly-created clients).
+	AgentClientUUID string
 }
 
 type clientScopeListItem struct {
@@ -71,6 +78,12 @@ func (a *Admin) EnsureAudienceScope(ctx context.Context, token string, p Audienc
 		return fmt.Errorf("verify audience mapper for scope %q: %w", scopeName, err)
 	}
 	_ = a.putRealmDefaultDefaultClientScope(ctx, token, p.Realm, scopeID)
+	// Attach the audience scope to the agent's own client. Required because Keycloak does
+	// not retroactively apply realm default-default-client-scopes to clients that already exist
+	// when the scope is added, and the agent's client is registered just before this call.
+	if p.AgentClientUUID != "" {
+		_ = a.putClientDefaultClientScope(ctx, token, p.Realm, p.AgentClientUUID, scopeID)
+	}
 	for _, plat := range p.PlatformClientIDs {
 		plat = strings.TrimSpace(plat)
 		if plat == "" {
--- a/kagenti-operator/internal/controller/clientregistration_controller.go
+++ b/kagenti-operator/internal/controller/clientregistration_controller.go
@@ -255,7 +255,7 @@ func (r *ClientRegistrationReconciler) reconcileOne(
 		logger.Error(err, "Keycloak admin token failed")
 		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
 	}
-	_, clientSecret, err := kc.RegisterOrFetchClientWithToken(ctx, token, keycloak.ClientRegistrationParams{
+	agentClientUUID, clientSecret, err := kc.RegisterOrFetchClientWithToken(ctx, token, keycloak.ClientRegistrationParams{
 		Realm:               ab.KeycloakRealm,
 		ClientID:            clientID,
 		ClientName:          clientName,
@@ -274,6 +274,7 @@ func (r *ClientRegistrationReconciler) reconcileOne(
 		AudienceClientID:     clientID,
 		PlatformClientIDs:    parsePlatformClientIDs(ab.PlatformClientIDs),
 		AudienceScopeEnabled: audienceScopeOn,
+		AgentClientUUID:      agentClientUUID,
 	}); err != nil {
 		logger.Error(err, "Keycloak audience scope management failed (credentials will still be written)",
 			"clientId", clientID)
--- a/kagenti-operator/internal/keycloak/audience_test.go
+++ b/kagenti-operator/internal/keycloak/audience_test.go
@@ -79,6 +79,127 @@ func TestEnsureAudienceScope(t *testing.T) {
 	}
 }
 
+// TestEnsureAudienceScope_AttachesToAgentClient verifies that when AgentClientUUID is set,
+// the audience scope is explicitly attached to the agent's own client as a default-client-scope
+// (in addition to the realm default-default registration). This is required because Keycloak
+// does not retroactively attach realm default-defaults to clients that already exist.
+func TestEnsureAudienceScope_AttachesToAgentClient(t *testing.T) {
+	var putAgentClientCalls, putPlatformClientCalls int
+	var srv *httptest.Server
+	srv = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		path := r.URL.Path
+		switch {
+		case path == testMasterRealmTokenPath:
+			w.Header().Set("Content-Type", "application/json")
+			_ = json.NewEncoder(w).Encode(map[string]string{"access_token": "tok"})
+		case path == "/admin/realms/kagenti/client-scopes" && r.Method == http.MethodGet:
+			_ = json.NewEncoder(w).Encode([]clientScopeListItem{})
+		case path == "/admin/realms/kagenti/client-scopes" && r.Method == http.MethodPost:
+			w.Header().Set("Location", srv.URL+"/admin/realms/kagenti/client-scopes/new-scope-id")
+			w.WriteHeader(http.StatusCreated)
+		case strings.Contains(path, "/client-scopes/new-scope-id/protocol-mappers/models") && r.Method == http.MethodPost:
+			w.WriteHeader(http.StatusCreated)
+		case strings.Contains(path, "/client-scopes/new-scope-id/protocol-mappers/models") && r.Method == http.MethodGet:
+			_ = json.NewEncoder(w).Encode([]protocolMapperRep{{
+				ID: "m1", Name: "agent-ns-wl-aud", Protocol: "openid-connect",
+				ProtocolMapper: "oidc-audience-mapper",
+				Config:         map[string]string{"included.custom.audience": "ns/wl"},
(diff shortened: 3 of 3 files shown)


mrsabath:
## Review feedback (non-blocking)

The fix is well-targeted. Root cause is accurately diagnosed (Keycloak's `default-default-client-scopes` only auto-attach to *newly-created* clients), and the fix piggybacks on the existing `putClientDefaultClientScope` helper rather than introducing a new code path. Three attachment sites now: realm-default, platform clients, agent's own client — symmetric and easy to follow. Tests follow the established `httptest` pattern in `audience_test.go`; the negative test's use of `t.Fatalf` in the default switch arm is the right way to assert *no* unexpected PUTs happen.

A few non-blocking recommendations:

1. **Best-effort error swallowing is consistent but invisible.** The new agent-client PUT discards the error with `_ =`, matching the existing pattern at `audience.go:73` and `:83`. Consistent — but if the agent attachment silently fails (transient 5xx, race with client deletion), the symptom is *exactly the bug being fixed*, with no log trace. Consider at least a `logger.V(1).Info("agent client scope attach failed", "err", err)` debug log so future regressions are diagnosable. Same argument applies to the two existing attachment sites; if you want to keep the change minimal here, a follow-up issue is fine.

2. **Consider an E2E assertion as a follow-up.** The PR was verified by manually applying the equivalent REST calls in-cluster. Once E2E is green, it'd be worth adding an assertion to the weather-agent demo test that the issued token's `aud` claim contains the SPIFFE ID — that would have caught this and would catch any future regression of the same shape.

3. **Doc comment nit on `AudienceParams.AgentClientUUID`.** The comment says "agents registered before the realm-level default-default scope existed won't have it attached." Accurate, but the more general failure mode is "any client created before the scope is registered as default-default" — which is *every* agent client given the current call ordering. Slight understatement of how universal the bug is. Not worth a re-push by itself.

CI status: all completed checks green (Build, CodeQL, Lint, Unit, Integration, Helm, Trivy, etc.). E2E and Verify Action Pinning still pending — worth waiting for those before merging.

mrsabath:
cc @akram — you wrote the original `EnsureAudienceScope` and the `included.custom.audience` mapper update. This PR adds explicit attachment of the audience scope to the agent's own client (the realm default-default mechanism doesn't apply to clients that already exist when the scope is registered). Would appreciate your eyes on the design choice.

kagenti-operator/internal/keycloak/audience.go
```diff
@@ -71,6 +78,12 @@ func (a *Admin) EnsureAudienceScope(ctx context.Context, token string, p Audienc
 		return fmt.Errorf("verify audience mapper for scope %q: %w", scopeName, err)
 	}
 	_ = a.putRealmDefaultDefaultClientScope(ctx, token, p.Realm, scopeID)
+	// Attach the audience scope to the agent's own client. Required because Keycloak does
+	// not retroactively apply realm default-default-client-scopes to clients that already exist
+	// when the scope is added, and the agent's client is registered just before this call.
+	if p.AgentClientUUID != "" {
```
pdettori, line 86: nit: Same `_ =` error-suppression as line 43 — both are best-effort. A `logger.V(1).Info("failed to attach audience scope to agent client", "uuid", p.AgentClientUUID, "err", err)` on failure would make field debugging easier without changing non-fatal semantics. Low priority since it follows the existing pattern.

pdettori (approved):
Clean, minimal fix for the Keycloak retroactivity issue — default-default-client-scopes don't auto-attach to pre-existing clients. The explicit PUT by UUID after registration is the correct solution. Tests cover both positive and backward-compatible empty-UUID cases. All CI green including E2E.

One minor nit inline (non-blocking).

akram:
> cc @akram — you wrote the original `EnsureAudienceScope` and the `included.custom.audience` mapper update. This PR adds explicit attachment of the audience scope to the agent's own client (the realm default-default mechanism doesn't apply to clients that already exist when the scope is registered). Would appreciate your eyes on the design choice.

@mrsabath thanks for this. This is indeed something that I missed in the initial `EnsureAudienceScope` implementation (`9637c03`).

The logic was always:

1. `putRealmDefaultDefaultClientScope` — register the scope as a realm default (expecting Keycloak to auto-attach it to all clients)
2. Explicit `putClientDefaultClientScope` loop — attach to each platform client

I assumed step 1 would cover the agent's own client, but Keycloak only auto-attaches `default-default-client-scopes` to clients created *after* the scope is registered. Since `RegisterOrFetchClientWithToken` runs before `EnsureAudienceScope` in `reconcileOne`, the agent's client already exists when the realm default is set — so it never gets the scope.

For platform clients, explicit attachment was already being done, but didn't apply the same reasoning to the agent's own client. The fix that you are proposing is correct: adding the agent client as a third explicit attachment site makes all three symmetric: realm-default (for future clients), platform clients, and agent client.

rhuss:
Clean, well-targeted fix. The root cause diagnosis is correct (Keycloak only auto-attaches `default-default-client-scopes` to newly-created clients, and the agent client is registered before `EnsureAudienceScope` runs). Making all three attachment sites symmetric (realm-default, platform clients, agent client) is the right approach, and the `if p.AgentClientUUID != ""` guard preserves backward compatibility cleanly.

Tests are solid: the positive case asserts the agent attachment happens, the negative case uses `t.Fatalf` in the default switch to catch any spurious PUTs. CI green.

One recommendation (non-blocking but worth doing before merge):

**Log the error instead of discarding it.** The `_ = a.putClientDefaultClientScope(...)` at the new agent attachment site means that if this call fails silently (transient 5xx, race with client deletion), the symptom is exactly the original bug with no log trace. A `logger.V(1).Info("agent client scope attach failed", "err", err)` would make regressions diagnosable. The same applies to the existing attachment sites at lines 73 and 83, but that can be a follow-up.

LGTM otherwise. The merge-to-main strategy (skip 0.6.x backport) makes sense given this is a first-boot race condition, not a steady-state bug.

mrsabath:
Good call — addressed in 76d9163. Replaced the silent `_ =` with a `log.FromContext(ctx).V(1).Info("agent client scope attach failed", ...)` carrying `clientUUID`, `scope`, and `err`, so a regression of the original bug would now leave a diagnostic trail.

Kept it scoped to just the new agent-attachment site for now; agree the existing realm/platform-client sites at lines 80 and 96 deserve the same treatment as a follow-up.

PR #396, opened 2026-06-02
Fix: remove kagenti.io/type label and config-hash on AgentRuntime deletion

rh-dnagornuks:
## Summary

When an AgentRuntime CR is deleted the controller previously preserved `kagenti.io/type` on the workload and replaced `kagenti.io/config-hash` with a defaults-only value, so the webhook continued to inject sidecars into replacement pods. This PR removes both the label and the annotation on deletion so that the next rolling update produces pods without sidecars, returning the workload to its pre-AR state. Docs and tests were updated to match the new lifecycle.

## Changes
- `handleDeletion`: remove `kagenti.io/type` from workload metadata labels and PodTemplateSpec pod labels instead of preserving it
- `handleDeletion`: delete `kagenti.io/config-hash` from PodTemplateSpec pod annotations instead of replacing it with a defaults-only hash; removes the now-unused `ComputeDefaultsOnlyHash` call
- Unit test: inverted assertions to confirm label and annotation are absent after deletion
- E2E `Agent lifecycle > should clean up on deletion`: updated to assert `kagenti.io/type` is removed from both metadata and PodTemplateSpec, and `config-hash` is absent
- E2E `Combined > should clean up on AgentRuntime deletion and stop injection` (renamed from "maintain injection"): updated to assert label/annotation removal and that replacement pods have no sidecars; replaced manual pod-delete-and-recheck loop with `WaitForDeploymentReady` + pod-list assertions
- `combined-agent` e2e fixture: removed `kagenti.io/type` from `spec.selector.matchLabels`, `spec.template.metadata.labels`, and `metadata.labels` — the controller owns this label's lifecycle; having it in the selector prevented the update during deletion (Kubernetes rejects removal of selector-required labels from the pod template)
- `docs/architecture.md`, `docs/controller-webhook-interaction.md`, `GETTING_STARTED.md`: updated deletion lifecycle descriptions

## Testing
- [x] `make test` and `pre-commit run --all-files` pass
- [x] On ROSA cluster, create AR targeting agent deployment and confirm sidecar injection
- [x] On ROSA cluster, delete AR and confirm that new pods don't receive sidecars

## Related Issues
Resolves: [RHAIENG-4933](https://issues.redhat.com/browse/RHAIENG-4933)

### DIFF
--- a/kagenti-operator/GETTING_STARTED.md
+++ b/kagenti-operator/GETTING_STARTED.md
@@ -170,8 +170,8 @@ spec:
 ### Deleting an AgentRuntime
 
 When you delete the AgentRuntime CR, the controller performs a graceful cleanup:
-- Preserves the `kagenti.io/type` label (so AgentCard discovery continues)
-- Updates the config hash to defaults-only (triggers a rollback to default sidecar configuration)
+- Removes the `kagenti.io/type` label from the workload metadata and PodTemplateSpec
+- Removes the `kagenti.io/config-hash` annotation from the PodTemplateSpec (triggers a rolling update so existing injected pods are replaced)
 - Removes the `app.kubernetes.io/managed-by` label
 
 ```bash
--- a/kagenti-operator/docs/architecture.md
+++ b/kagenti-operator/docs/architecture.md
@@ -72,7 +72,7 @@ The Kagenti Operator is a Kubernetes controller that implements the [Operator Pa
 - Computes config hash from 3-layer merged configuration (cluster defaults → namespace defaults → CR overrides)
 - Discovers linked skills by reading the `kagenti.io/skills` annotation from target workloads when the `skillDiscovery` feature gate is enabled
 - Triggers rolling updates when configuration changes
-- On CR deletion: preserves type label, updates config-hash to defaults-only, removes managed-by label
+- On CR deletion: removes type label, managed-by label and config-hash annotation (causing the workload to lose sidecars)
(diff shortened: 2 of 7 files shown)


pdettori (approved):
Clean, well-structured PR. The deletion logic change is sound — removing `kagenti.io/type` instead of preserving it correctly ensures the webhook pre-filter skips future pods. The `delete()` calls on maps are idempotent and safe. The fixture change (removing `kagenti.io/type` from `spec.selector.matchLabels`) is well-motivated — Kubernetes rejects updates that remove a selector-required label from the pod template (422), which would cause the finalizer to never clear.

**Areas reviewed**: Go controller logic, unit tests, E2E tests, fixtures, docs, security
**Commits**: 3 commits, all signed-off ✓
**CI status**: All passing (E2E pending at review time)

**Nit** (non-blocking): Commit `1efe0fd` uses `Fix:` (capitalized) prefix instead of `fix:` (lowercase conventional commit). Minor inconsistency with the other two commits.

pdettori:
The E2E failure is unrelated to this PR. The failed test is `Skill Discovery E2E [Feature gate enabled] should populate linkedSkills from annotation` — it timed out (180s) waiting for `status.linkedSkills` to contain `"summarizer"` but the field stayed empty. This is a pre-existing flaky/infrastructure issue with the skill discovery feature gate mechanism, not caused by the deletion lifecycle changes here.

PR #397, opened 2026-06-02
feat: operator auto-discovers and manages identity config

ChristianZaccaria:
## Summary
Shifts identity bootstrapping from the setup script/Helm chart to the operator. The operator now reads keycloak-initial-admin directly, auto-discovers Keycloak public URL and SPIRE trust domain, and reconciles authbridge-config ConfigMaps in labeled agent namespaces.

## Changes
- Add authbridge-config reconciler — creates/updates ConfigMap in namespaces with `kagenti-enabled=true`
- Add auto-discovery package — discovers Keycloak URL from Route/Ingress, SPIRE trust domain from ZTWIM CR/spire-bundle
- Remove legacy keycloak-admin-secret fallback from ClientRegistrationReconciler (reads keycloak-initial-admin directly)
- Default `--enable-authbridge-config=true` in binary
- Add RBAC for Routes, Ingresses, ZeroTrustWorkloadIdentityManagers (Helm + Kustomize)
- Add Kustomize webhook namespace selector patch (prevents chicken-and-egg pod creation deadlock)

## Test Plan
- [ ] Deploy operator via Helm, verify auto-discovery logs: "Auto-discovered Keycloak public URL" and "Auto-discovered SPIRE trust domain"
- [ ] Label a namespace `kagenti-enabled=true`, verify authbridge-config ConfigMap is created with correct values
- [ ] Delete the ConfigMap, verify self-healing (recreated)
- [ ] Send A2A message to weather agent, verify response (confirms direct secret reading + client registration)
- [ ] Delete operator pod, verify it restarts (webhook namespace selector fix)

### DIFF
--- a/charts/kagenti-operator/values.yaml
+++ b/charts/kagenti-operator/values.yaml
@@ -107,6 +107,8 @@ keycloak:
 authbridgeConfig:
   enabled: true
   clientAuthType: "client-secret"
+  # How long AuthBridge waits for Keycloak credentials (increase for slow Keycloak startup)
+  # credentialWaitTimeout: "120s"
 
 # [MLFLOW]: MLflow experiment tracking integration
 mlflow:
--- a/charts/kagenti-operator/values.yaml
+++ b/charts/kagenti-operator/values.yaml
@@ -92,6 +92,22 @@ networkPolicy:
 # Works with both verifiedFetch (mTLS) and signatureVerification (init-signer).
 enforceNetworkPolicies: false
 
+# [KEYCLOAK]: Keycloak integration settings
+keycloak:
+  # Namespace where the keycloak-initial-admin secret is located
+  adminSecretNamespace: keycloak
+  # Target realm for agent client registration
+  realm: kagenti
+  # External Keycloak URL (used for ISSUER / EXPECTED_AUDIENCE in authbridge-config).
+  # Auto-discovered from OpenShift Route or Kubernetes Ingress named "keycloak" in
+  # the adminSecretNamespace. Set here only to override auto-discovery.
+  publicUrl: ""
+
+# [AUTHBRIDGE CONFIG]: Reconcile authbridge-config ConfigMap in agent namespaces
+authbridgeConfig:
+  enabled: true
+  clientAuthType: "client-secret"
+
 # [MLFLOW]: MLflow experiment tracking integration
 mlflow:
   enable: false
@@ -112,8 +128,6 @@ verifiedFetch:
   spiffeEndpointSocket: "unix:///spiffe-workload-api/spire-agent.sock"
   # Agents must expose a Service port named "agent-tls" (default 8443).
   # The operator discovers the TLS endpoint by port name, not number.
-  # SPIRE trust domain for identity binding (required when enabled)
-  spireTrustDomain: ""
 
 # [SIGNATURE VERIFICATION]: A2A agent card signature verification via SPIRE x5c
 signatureVerification:
@@ -123,12 +137,9 @@ signatureVerification:
   auditMode: false
   # Enforce network policies based on signature verification
   enforceNetworkPolicies: false
-  # SPIRE trust domain. Required when SPIRE is enabled in the cluster even if
-  # signatureVerification.enabled is false: the ClientRegistrationReconciler
-  # uses this to build SPIFFE-shaped Keycloak client IDs
-  # (spiffe://<trust-domain>/ns/<ns>/sa/<sa>) whenever authbridge-config has
-  # SPIRE_ENABLED=true. The chart renders --spire-trust-domain independently of
-  # signatureVerification.enabled for this reason.
+  # SPIRE trust domain. Auto-discovered from ZTWIM CR or spire-bundle ConfigMap.
+  # Set here only to override auto-discovery. Used by ClientRegistrationReconciler
+  # to build SPIFFE-shaped Keycloak client IDs (spiffe://<trust-domain>/ns/<ns>/sa/<sa>).
   spireTrustDomain: ""
   # Key within the SPIRE trust bundle ConfigMap. Matches the SPIRE hardened Helm chart default
   # and the binary flag default. Override to "bundle.crt" only for older ZTWIM deployments.
--- a/kagenti-operator/config/manager/kustomization.yaml
+++ b/kagenti-operator/config/manager/kustomization.yaml
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/rh_ee_czaccari/kagenti-operator
-  newTag: testmychanges4
+  newName: kagenti-operator
+  newTag: v0.0.1
--- a/kagenti-operator/config/manager/kustomization.yaml
+++ b/kagenti-operator/config/manager/kustomization.yaml
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: kagenti-operator
-  newTag: v0.0.1
+  newName: quay.io/rh_ee_czaccari/kagenti-operator
+  newTag: testmychanges4
--- a/kagenti-operator/internal/controller/authbridgeconfig_controller.go
+++ b/kagenti-operator/internal/controller/authbridgeconfig_controller.go
@@ -36,6 +36,7 @@ type AuthbridgeConfigPlatform struct {
 	SpireTrustDomain             string
 	ClientAuthType               string
 	SpiffeIdpAlias               string
+	CredentialWaitTimeout        string
 }
 
 // AuthbridgeConfigReconciler reconciles authbridge-config ConfigMaps in namespaces
@@ -160,7 +161,7 @@ func (r *AuthbridgeConfigReconciler) buildConfigMapData() map[string]string {
 		"SPIRE_ENABLED":           spireEnabled,
 		"CLIENT_AUTH_TYPE":        clientAuthType,
 		"SPIFFE_IDP_ALIAS":        spiffeIdpAlias,
-		"CREDENTIAL_WAIT_TIMEOUT": "120s",
+		"CREDENTIAL_WAIT_TIMEOUT": r.credentialWaitTimeout(),
 	}
 
 	if issuer != "" {
@@ -170,6 +171,13 @@ func (r *AuthbridgeConfigReconciler) buildConfigMapData() map[string]string {
 	}
 
 	return data
+}
+
+func (r *AuthbridgeConfigReconciler) credentialWaitTimeout() string {
+	if r.Platform.CredentialWaitTimeout != "" {
+		return r.Platform.CredentialWaitTimeout
+	}
+	return "120s"
 }
 
 func (r *AuthbridgeConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
--- a/kagenti-operator/internal/controller/authbridgeconfig_controller.go
+++ b/kagenti-operator/internal/controller/authbridgeconfig_controller.go
@@ -1,0 +1,195 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"fmt"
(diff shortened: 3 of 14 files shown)


kagenti-operator/config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: kagenti-operator
-  newTag: v0.0.1
+  newName: quay.io/rh_ee_czaccari/kagenti-operator
+  newTag: testmychanges4
```
pdettori: **must-fix:** Test image `quay.io/rh_ee_czaccari/kagenti-operator:testmychanges4` must be reverted to the standard image before merge. This looks like a local dev override left in.

kagenti-operator/internal/discovery/discovery.go
```diff
@@ -0,0 +1,144 @@
...
+		return url, nil
+	}
+
+	return "", fmt.Errorf("could not discover Keycloak public URL from Route or Ingress in namespace %q", keycloakNamespace)
+}
+
+func discoverFromRoute(ctx context.Context, c client.Reader, ns string) (string, error) {
+	route := &unstructured.Unstructured{}
+	route.SetGroupVersionKind(routeGVK)
+	if err := c.Get(ctx, types.NamespacedName{Namespace: ns, Name: "keycloak"}, route); err != nil {
```
pdettori, line 53: **suggestion:** The discovery functions log at Info level and return empty string on failure. Consider returning an explicit error so callers can distinguish "not found" from "permission denied" (e.g., missing RBAC for Routes on non-OpenShift clusters).

kagenti-operator/internal/controller/authbridgeconfig_controller.go
```diff
@@ -0,0 +1,195 @@
...
+				}
+			} else {
+				logger.Error(createErr, "Failed to create authbridge-config", "namespace", ns.Name)
+				return ctrl.Result{}, createErr
+			}
+		} else {
+			logger.Info("Created authbridge-config", "namespace", ns.Name)
+			return ctrl.Result{}, nil
+		}
+	} else if err != nil {
```
pdettori, line 101: **suggestion:** `CREDENTIAL_WAIT_TIMEOUT` is hardcoded to `"120s"`. Consider making this configurable via values/flag for environments with slow Keycloak startup.

pdettori (changes requested):
## Review Summary

Well-structured PR that moves identity bootstrapping into the operator with proper auto-discovery fallbacks. The architecture is sound — ConfigMap reconciliation in labeled namespaces follows the operator pattern correctly, and the migration from `keycloak-admin-secret` to `keycloak-initial-admin` (RHBK-native) is a good alignment with upstream.

**Two must-fix issues** block merge: a leftover test image in kustomization.yaml and the E2E test failures from the removed CLI flag.

**Areas reviewed:** Go code, Helm, Kustomize, RBAC, CI failure analysis
**Commits:** 1 commit, signed-off: yes
**CI status:** E2E failing (PR-introduced, fixable)

---

### CI Failure Root Cause

The PR removes the `--spire-trust-domain` CLI flag (replaced by auto-discovery + env var), but E2E tests at lines 851 and 1511 of `e2e_test.go` still inject this flag via `utils.PatchControllerArgs()`. Fix: update tests to patch the env var or remove the flag injection entirely.

kagenti-operator/internal/controller/authbridgeconfig_controller.go
ChristianZaccaria: The timeout is now configurable via:

- CLI flag: `--credential-wait-timeout=120s`
- Helm value: `authbridgeConfig.credentialWaitTimeout: "120s"`
- Defaults to "120s" if not set

kagenti-operator/config/manager/kustomization.yaml
ChristianZaccaria: Whoops, thank you for spotting this!

pdettori (commented):
Thanks for addressing the test image and configurable timeout — both look good.

The E2E failure is still the same root cause from my original review: tests at `e2e_test.go:855` and `e2e_test.go:1516` patch the controller with `--spire-trust-domain=example.org`, but that flag no longer exists. The manager crashes with `flag provided but not defined: -spire-trust-domain` and the deployment never rolls out.

**Fix**: update those test helpers to set `KAGENTI_SPIRE_TRUST_DOMAIN` as an env var on the deployment instead of passing the removed CLI flag.

Once that's fixed, happy to approve.

pdettori:
## E2E Test Failure — Root Cause Analysis

**Failed test**: `AgentCard E2E [With signature verification] [BeforeAll] Audit mode - should allow sync but report SignatureInvalidAudit`

### Root Cause: Two-step deployment mutation causes a restart race

The `BeforeAll` setup performs two sequential operations that each trigger a pod restart:

1. `PatchControllerArgs(...)` with `--require-a2a-signature=true` → triggers rollout immediately
2. `kubectl set env ... KAGENTI_SPIRE_TRUST_DOMAIN=example.org` → intended to provide the trust domain

The problem: Step 1 triggers a restart **before** Step 2 runs. The new pod starts with `--require-a2a-signature=true` but has no `KAGENTI_SPIRE_TRUST_DOMAIN` env var yet. Auto-discovery also fails because there's no `spire-bundle` ConfigMap in the `zero-trust-workload-identity-manager` namespace (the test references `spire-system`).

The controller hits the hard exit:
```
spireTrustDomain is required when --require-a2a-signature=true
(set KAGENTI_SPIRE_TRUST_DOMAIN or ensure ZTWIM/spire-bundle is accessible)
os.Exit(1)
```

The pod enters CrashLoopBackOff, the old replica never terminates, and `WaitForRollout` times out at 2 minutes.

### Suggested Fix

Combine both mutations into a single `kubectl patch` so the deployment restarts only once with all required config present:

```go
// Single atomic patch — args + env var together
patch := `{"spec":{"template":{"spec":{"containers":[{"name":"manager","env":[{"name":"KAGENTI_SPIRE_TRUST_DOMAIN","value":"example.org"}]}]}}}}`
cmd := exec.Command("kubectl", "patch", "deployment/"+controllerDeployment,
    "-n", controllerNamespace, "--type=strategic", "-p", patch)
```

Combined with the existing `PatchControllerArgs` call (or merged into one patch), this ensures the pod never starts in an invalid intermediate state.

The same pattern applies to the other `BeforeAll` block (around line 876 in the new diff) that also sets `KAGENTI_SPIRE_TRUST_DOMAIN` separately.

---
Assisted-By: Claude Code

ChristianZaccaria:
@pdettori thank you for the suggested fix and for looking into this.

The E2E test was crashing because it patched the controller in two steps, first adding the signature flag, then setting the trust domain env var. The pod restarted after the first step before the second ran, so it had no trust domain and exited immediately. Now both are applied together so the pod always starts with everything it needs.

kagenti-operator/internal/controller/authbridgeconfig_controller.go
```diff
@@ -0,0 +1,203 @@
...
+
+	spiffeIdpAlias := r.Platform.SpiffeIdpAlias
+	if spiffeIdpAlias == "" {
+		spiffeIdpAlias = "spire-spiffe"
+	}
+
+	data := map[string]string{
+		"KEYCLOAK_URL":            keycloakURL,
+		"KEYCLOAK_REALM":          realm,
+		"KEYCLOAK_NAMESPACE":      kcNS,
```
pdettori, line 160: suggestion: The ConfigMap sets `SPIRE_ENABLED=true` when `SpireTrustDomain != ""`, but the trust domain value itself isn't included in the CM data. The `ClientRegistrationReconciler` reads `SPIRE_ENABLED` from this CM to decide whether to use SPIFFE-shaped client IDs — is the trust domain propagated to the webhook sidecars via another mechanism? If so, a brief comment here ("trust domain passed via X") would make the data flow easier to follow.

kagenti-operator/internal/controller/authbridgeconfig_controller.go
```diff
@@ -0,0 +1,203 @@
...
+	cm.Labels["app.kubernetes.io/managed-by"] = "kagenti-operator"
+	cm.Labels[LabelNamespaceDefaults] = "true"
+	if cm.Data == nil {
+		cm.Data = make(map[string]string)
+	}
+	for k, v := range desired {
+		cm.Data[k] = v
+	}
+	if updateErr := r.Update(ctx, cm); updateErr != nil {
+		logger.Error(updateErr, "Failed to update authbridge-config", "namespace", ns.Name)
```
pdettori, line 118: nit: The reconciler always calls `Update()` even when `cm.Data` already matches `desired`. A `reflect.DeepEqual(cm.Data, desired)` short-circuit before the update would avoid unnecessary writes and downstream reconcile churn from watchers.

charts/kagenti-operator/values.yaml
```diff
@@ -92,6 +92,24 @@ networkPolicy:
...
+  # Namespace where the keycloak-initial-admin secret is located
+  adminSecretNamespace: keycloak
+  # Target realm for agent client registration
+  realm: kagenti
+  # External Keycloak URL (used for ISSUER / EXPECTED_AUDIENCE in authbridge-config).
+  # Auto-discovered from OpenShift Route or Kubernetes Ingress named "keycloak" in
+  # the adminSecretNamespace. Set here only to override auto-discovery.
+  publicUrl: ""
+
+# [AUTHBRIDGE CONFIG]: Reconcile authbridge-config ConfigMap in agent namespaces
```
pdettori, line 106: suggestion: Both the binary default (`flag.BoolVar(..., true)`) and the values.yaml default (`enabled: true`) enable authbridge-config. The template only emits `--enable-authbridge-config=false` when this is explicitly set to false. Consider a brief inline comment noting the binary default to help operators reading values.yaml in isolation.

kagenti-operator/internal/discovery/discovery.go
```diff
@@ -0,0 +1,150 @@
...
+// ConfigMap which contains the trust domain as a JSON key in SPIFFE bundle format.
+func DiscoverSPIRETrustDomain(ctx context.Context, c client.Reader, spireNamespace string) (string, error) {
+	td, ztwimErr := discoverFromZTWIM(ctx, c)
+	if ztwimErr == nil && td != "" {
+		return td, nil
+	}
+
+	td, bundleErr := discoverFromSpireBundle(ctx, c, spireNamespace)
+	if bundleErr == nil && td != "" {
+		return td, nil
```
pdettori, line 85: suggestion: Consider logging which CM key was matched (e.g. "Discovered trust domain from spire-bundle key 'bundle.spiffe'"). In the field, debugging misconfigured SPIFFE bundles is much easier when the operator logs say which key resolved rather than just the resulting trust domain.

kagenti-operator/test/utils/utils.go
```diff
@@ -401,9 +401,44 @@ func buildArgsPatch(argsJSON []byte) string {
...
+				"spec": map[string]interface{}{
+					"containers": []container{
+						{Name: "manager", Args: args, Env: envList},
+					},
+				},
+			},
+		},
+	}
+	out, _ := json.Marshal(p)
+	return string(out), "strategic"
```
pdettori, line 434: nit (non-blocking): `json.Marshal` errors are silently discarded. Since the input is fully controlled in tests this is fine, but a brief `// input is static; marshal cannot fail` comment (or wrapping in a must-helper) clarifies intent.

pdettori (approved):
Well-architected PR that moves identity bootstrapping from imperative scripting into a declarative operator reconciliation loop. The auto-discovery chain (ZTWIM CR → spire-bundle CM → env var → fallback) is well-ordered and the E2E test migration from `--spire-trust-domain` flag to `KAGENTI_SPIRE_TRUST_DOMAIN` env var is correctly implemented. The atomic strategic-merge patch for E2E test setup eliminates the prior race between separate arg and env patches.

All 15 CI checks pass. No hardcoded secrets. RBAC additions are properly scoped. The webhook namespace selector patch correctly prevents the chicken-and-egg deadlock.

Previous CHANGES_REQUESTED feedback (test image, E2E flag migration) has been addressed.

**Areas reviewed**: Go, Helm, Kustomize, RBAC, E2E tests, CI status
**Commits**: 4 commits, all signed-off
**CI status**: all passing

PR #402, opened 2026-06-04
feat: remove enableTracing and tracingBackend from platform config

Bobbins228:
## Summary

Removes the `observability.enableTracing` and `observability.tracingBackend` fields from the kagenti-platform-config schema. Tracing is deferred and MLflow connection is now based on MLF CR presence, so these fields are dangling configuration with no consumer.

The YAML unmarshaller silently ignores unknown keys, so existing ConfigMaps that still carry these fields continue to load without error (backward compat during upgrade).

## Changes

- `internal/webhook/config/types.go` — removed `EnableTracing` and `TracingBackend` fields from `ObservabilityConfig` struct
- `internal/webhook/config/defaults.go` — removed `EnableTracing: false` from compiled defaults

## Test Plan

- [x] `go build ./...` passes
- [x] `go test ./internal/webhook/config/...` passes
- [x] `rg` confirms zero remaining references to removed fields
- [ ] CI passes (e2e requires cluster)

### DIFF
--- a/kagenti-operator/internal/webhook/config/defaults.go
+++ b/kagenti-operator/internal/webhook/config/defaults.go
@@ -75,7 +75,6 @@ func CompiledDefaults() *PlatformConfig {
 		Observability: ObservabilityConfig{
 			LogLevel:      "info",
 			EnableMetrics: true,
-			EnableTracing: false,
 		},
 	}
 }
--- a/kagenti-operator/internal/webhook/config/types.go
+++ b/kagenti-operator/internal/webhook/config/types.go
@@ -65,10 +65,8 @@ type SpiffeConfig struct {
 }
 
 type ObservabilityConfig struct {
-	LogLevel       string `json:"logLevel" yaml:"logLevel"`
-	EnableMetrics  bool   `json:"enableMetrics" yaml:"enableMetrics"`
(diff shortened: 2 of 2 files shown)


r3v5 (approved):
lgtm

cwiklik (approved):
Clean removal of two dangling `ObservabilityConfig` fields (`EnableTracing`, `TracingBackend`) that have no consumer now that tracing is deferred and MLflow wiring is CR-presence-based.

Verified safe:
- **Zero consumers** — org-wide code search finds these fields referenced only in `types.go` itself; no Go reader and no chart/values in the org emits the keys. Build + Unit Tests green confirm nothing compiles against them.
- **Backward-compat holds** — no `UnmarshalStrict`/`DisallowUnknownFields` anywhere, so the loader is lenient and existing ConfigMaps still carrying the old keys load without error during upgrade, exactly as the PR states.

**Optional, non-blocking:** a tiny regression test asserting a ConfigMap with the legacy keys still unmarshals cleanly would lock in the lenient-parsing guarantee this change relies on (cheap insurance if strict unmarshalling is ever enabled). No dead-config cleanup is owed — no org chart emits these keys.

**Commits:** signed-off. Nit: `feat:` is arguably `refactor:`/`chore:` for a removal, and lacks the emoji prefix the convention favors (style only).

All 16 CI checks passing.

PR #404, opened 2026-06-04
Fix: use correct MLflow API group and resource for agent RBAC

rh-dnagornuks:
## Summary

Fixes the MLflow integration constants to use `mlflow.kubeflow.org` (API group) and `experiments` (resource) instead of the incorrect `mlflow.opendatahub.io` / `mlflowexperiments`. The MLflow gateway's Kubernetes authorization plugin performs SubjectAccessReview checks against `mlflow.kubeflow.org/experiments`, not the CRD's own API group — so agents were getting 403 PERMISSION_DENIED when exporting traces via the MLflow SDK.

This is a follow-up to #366. Further testing revealed that the Role permissions weren't properly configured, possibly because traces exported by the OTel collector masked failures in the agent's direct MLflow SDK path.

## Changes
- `MLflowExperimentsAPIGroup`: `mlflow.opendatahub.io` → `mlflow.kubeflow.org`
- `MLflowExperimentsResource`: `mlflowexperiments` → `experiments`
- Updated kubebuilder RBAC markers to match
- Updated `config/rbac/role.yaml` and `charts/kagenti-operator/templates/rbac/role.yaml`
- Updated `docs/mlflow-integration.md` to reflect correct values

## Testing
- Deployed an [agent](https://quay.io/repository/dnagornu_org/generic-agent) with MLFlow tracing enabled in code on a fresh ROSA cluster
- Removed `OTEL_EXPORTER_OTLP_ENDPOINT` to isolate the MLflow SDK path
- Verified `POST /mlflow/api/3.0/mlflow/traces` returns 200 with full span data (4 nested spans, token counts)

### DIFF
--- a/charts/kagenti-operator/templates/rbac/role.yaml
+++ b/charts/kagenti-operator/templates/rbac/role.yaml
@@ -169,9 +169,9 @@ rules:
   - update
   - watch
 - apiGroups:
-  - mlflow.opendatahub.io
+  - mlflow.kubeflow.org
   resources:
-  - mlflowexperiments
+  - experiments
   verbs:
   - get
   - list
--- a/kagenti-operator/config/rbac/role.yaml
+++ b/kagenti-operator/config/rbac/role.yaml
@@ -141,9 +141,9 @@ rules:
   - list
   - watch
 - apiGroups:
-  - mlflow.opendatahub.io
+  - mlflow.kubeflow.org
   resources:
-  - mlflowexperiments
+  - experiments
   verbs:
   - get
   - list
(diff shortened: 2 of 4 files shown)


cwiklik (approved):
Correct, well-scoped fix: the per-agent Role must grant on `mlflow.kubeflow.org/experiments` (what the MLflow gateway's authorization plugin checks via SubjectAccessReview), not the CRD's own `mlflow.opendatahub.io/mlflowexperiments` — which is why agents hit 403 on the direct MLflow SDK trace path.

Verified thorough:
- **Complete & consistent** — updates the Go constants, the kubebuilder marker, **both** `config/rbac/role.yaml` and the chart `role.yaml` (correct 3-way RBAC sync), and the doc. The `mlflows` marker correctly stays on `mlflow.opendatahub.io` (the real CRD); only the gateway's virtual `experiments` resource moves to `mlflow.kubeflow.org`.
- **Regression-protected** — `mlflow_controller_test.go:244-245` asserts the generated Role uses the constants, so the corrected values are locked in (and explains the green Unit Tests without editing the test).
- **Empirically validated** — author confirmed `POST /mlflow/api/3.0/mlflow/traces` returns 200 with full spans on a fresh ROSA cluster after isolating the SDK path.

**CI — E2E red but unrelated:** the one failure is `SharedTrust E2E [BeforeAll] should create cacerts Secrets` — timed out 120s (32/33 passed); an Istio shared-trust setup flake with no connection to MLflow RBAC constants. Recommend a re-run before merge.

**Security:** tightens correctness of least-privilege agent RBAC; scoped via `resourceNames`. No secrets.

**Commits:** signed-off. Nit: subject `Fix:` → conventional `fix:` (lowercase), lacks emoji prefix (style only).

pdettori:
@rh-dnagornuks CI has been fixed - please rebase from upstream main and force push so we can check now with full CI

PR #406, opened 2026-06-05
feat: Wire proxy-sidecar egress enforcement (UID fix + enforce-drop)

huang195:
## Summary

Wires the proxy-sidecar **fail-closed egress guard** into the operator. `proxy-sidecar` mode routes egress through AuthBridge via `HTTP_PROXY`, which is **cooperative** — an app that ignores `HTTP_PROXY` (or sets `NO_PROXY`) egresses directly and bypasses AuthBridge. [kagenti-extensions#484](https://github.com/kagenti/kagenti-extensions/pull/484) (merged) added `MODE=enforce-drop` to the shared `proxy-init` image; this PR injects it (opt-in, default off) plus the prerequisite UID fix.

This is the operator half of the proxy-sidecar egress-enforcement work (extensions#484 was the image half).

## What changed

- **UID fix (unconditional):** the proxy-sidecar `authbridge-proxy` container hardcoded `RunAsUser: 1001` — the same UID as the app — so the enforce-drop `--uid-owner` exemption couldn't distinguish proxy from app egress. Now derived from `cfg.Proxy.UID` (1337), the single source of truth also used for the proxy-init `PROXY_UID`, matching the envoy builder. Adds `RunAsGroup` parity and validates `Proxy.UID >= 1`.
- **Config (`PlatformConfig.Proxy`):** `EgressEnforcement` (`off`|`enforce`, default `off`) and `ClusterCIDRs` (default `10.0.0.0/8`). Surfaced via the chart's `defaults.proxy`.
- **Injection:** `BuildProxyInitContainer` gains a `mode` param — `redirect` (envoy-sidecar, unchanged) or `enforce-drop` (emits `MODE` + `CLUSTER_CIDRS`, omits the redirect-only `POD_IP`/exclude vars). The proxy-sidecar branch injects it when `EgressEnforcement=="enforce"`.

## Design notes

- **Default off → keeps `main` shippable.** No behavior change until an operator sets `enforce`. The end goal is enforce-by-default: a later release flips the *compiled default* `off → enforce` (one line) once `ClusterCIDRs` sourcing is validated across distros. The knob is a permanent operational override — **nothing to remove**, no throwaway gate.
- **envoy-sidecar is unaffected.** It already enforces egress structurally via the transparent redirect, so the knob is consulted only on the proxy-sidecar / lite path — no mode-compatibility matrix.
- **Security:** the guard rests on UID separation (the agent must not run as, or `setuid()` to, `Proxy.UID`). `Proxy.UID >= 1` is validated. Before flipping the default to `enforce`, confirm PSA `restricted` / SCC enforces non-root + no `CAP_SETUID` on agents (or add a `runAsUser == Proxy.UID` admission check).

## Rollout notes / dependencies

- **`ClusterCIDRs` default is Kind-shaped** (`10.0.0.0/8`). **OCP/EKS must override** (e.g. OCP services `172.30.0.0/16` — outside `10/8` — and pods `10.128.0.0/14`) or in-cluster service traffic is dropped. Documented in `values.yaml` and the field doc.
- `enforce-drop` intentionally ignores `OUTBOUND_PORTS_EXCLUDE` (redirect-only); anything previously bypassed that way must route through the proxy or fall in `ClusterCIDRs`.
- Release coordination: the install chart must deploy a `proxy-init` image containing extensions#484. Image versions are owned by the install chart, not pinned here.

## Testing

`go build ./... && go test ./internal/webhook/...` — all green. New tests: proxy container `RunAsUser == Proxy.UID`; proxy-init `enforce-drop` env (`MODE`/`PROXY_UID`/`CLUSTER_CIDRS`, no `POD_IP`); injection present on `enforce` / absent on default `off` / not duplicated. Helm `defaults.proxy` renders the new fields.

### DIFF
--- a/kagenti-operator/internal/webhook/config/types.go
+++ b/kagenti-operator/internal/webhook/config/types.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"net"
 
 	corev1 "k8s.io/api/core/v1"
 )
@@ -159,6 +160,21 @@ func (c *PlatformConfig) Validate() error {
 	case "", EgressEnforcementOff, EgressEnforcementEnforce:
 	default:
 		return fmt.Errorf("proxy.egressEnforcement must be \"off\" or \"enforce\" (got %q)", c.Proxy.EgressEnforcement)
+	}
+	// When enforce is on, ClusterCIDRs drive the only in-cluster allowance in the
+	// enforce-drop guard. Validate them at load time so a misconfig fails fast with
+	// a clear message rather than: (a) an empty list silently falling back to the
+	// Kind-shaped 10.0.0.0/8 default in init-iptables.sh, or (b) a malformed entry
+	// crashing the proxy-init container under `set -e` with a cryptic iptables error.
+	if c.Proxy.EgressEnforcement == EgressEnforcementEnforce {
+		if len(c.Proxy.ClusterCIDRs) == 0 {
+			return fmt.Errorf("proxy.clusterCIDRs must be non-empty when proxy.egressEnforcement is %q (set the cluster's pod+service CIDRs)", EgressEnforcementEnforce)
+		}
+		for _, cidr := range c.Proxy.ClusterCIDRs {
+			if _, _, err := net.ParseCIDR(cidr); err != nil {
+				return fmt.Errorf("proxy.clusterCIDRs entry %q is not a valid CIDR: %w", cidr, err)
+			}
+		}
 	}
 	if c.Images.EnvoyProxy == "" {
 		return fmt.Errorf("images.envoyProxy is required")
--- a/kagenti-operator/internal/webhook/config/types.go
+++ b/kagenti-operator/internal/webhook/config/types.go
@@ -170,6 +170,9 @@ func (c *PlatformConfig) Validate() error {
 		if len(c.Proxy.ClusterCIDRs) == 0 {
 			return fmt.Errorf("proxy.clusterCIDRs must be non-empty when proxy.egressEnforcement is %q (set the cluster's pod+service CIDRs)", EgressEnforcementEnforce)
 		}
+		// Syntactic validation only — overlapping ranges and IPv4/IPv6 mixing are
+		// accepted (iptables handles both, and the init script splits v4/v6 itself);
+		// we only reject malformed strings here.
 		for _, cidr := range c.Proxy.ClusterCIDRs {
 			if _, _, err := net.ParseCIDR(cidr); err != nil {
 				return fmt.Errorf("proxy.clusterCIDRs entry %q is not a valid CIDR: %w", cidr, err)
--- a/kagenti-operator/internal/webhook/config/types.go
+++ b/kagenti-operator/internal/webhook/config/types.go
@@ -40,11 +40,37 @@ type ImageConfig struct {
 	PullPolicy corev1.PullPolicy `json:"pullPolicy" yaml:"pullPolicy"`
 }
 
+// EgressEnforcement values for ProxyConfig.EgressEnforcement.
+const (
+	EgressEnforcementOff     = "off"
+	EgressEnforcementEnforce = "enforce"
+)
+
 type ProxyConfig struct {
 	Port             int32 `json:"port" yaml:"port"`
 	UID              int64 `json:"uid" yaml:"uid"`
 	InboundProxyPort int32 `json:"inboundProxyPort" yaml:"inboundProxyPort"`
 	AdminPort        int32 `json:"adminPort" yaml:"adminPort"`
+
+	// EgressEnforcement controls the proxy-sidecar fail-closed egress guard.
+	//   "off" (default): the workload routes egress through the forward proxy
+	//     via HTTP_PROXY only — cooperative and bypassable.
+	//   "enforce": a proxy-init container installs the enforce-drop iptables
+	//     guard, forcing all external egress through the forward proxy regardless
+	//     of whether the workload honors HTTP_PROXY.
+	// envoy-sidecar mode is unaffected — it already enforces egress structurally
+	// via the transparent redirect, so this knob is consulted only on the
+	// proxy-sidecar / lite path. Migrate the default off->enforce in a future
+	// release once ClusterCIDRs sourcing is validated across distros.
+	EgressEnforcement string `json:"egressEnforcement" yaml:"egressEnforcement"`
+
+	// ClusterCIDRs are the in-cluster ranges (pods / services / DNS) that the
+	// enforce-drop guard allows direct; everything else egressing the pod is
+	// dropped. The default 10.0.0.0/8 is Kind-shaped (pods 10.244/16 + services
+	// 10.96/16). OCP/EKS MUST override this (e.g. OCP services 172.30.0.0/16,
+	// pods 10.128.0.0/14 — 172.30/16 is outside 10/8) or in-cluster service
+	// traffic will be dropped. Only used when EgressEnforcement == "enforce".
+	ClusterCIDRs []string `json:"clusterCIDRs" yaml:"clusterCIDRs"`
 }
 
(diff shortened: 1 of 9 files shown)


kagenti-operator/internal/webhook/injector/container_builder.go
```diff
@@ -448,51 +454,72 @@ func (b *ContainerBuilder) buildEnvoyProxyEnvLegacy() []corev1.EnvVar {
...
+//   - "redirect" (envoy-sidecar): transparently REDIRECT pod traffic to the
+//     Envoy listeners. outboundPortsExclude / inboundPortsExclude (from the
+//     kagenti.io/outbound-ports-exclude and kagenti.io/inbound-ports-exclude
+//     pod annotations; mandatory 8080 always included outbound) tune it.
+//   - "enforce-drop" (proxy-sidecar): a fail-closed egress guard that DROPs any
+//     egress bypassing the forward proxy. Driven by PROXY_UID + CLUSTER_CIDRS;
+//     the exclude args do not apply (the script ignores them in this mode).
+func (b *ContainerBuilder) BuildProxyInitContainer(mode, outboundPortsExclude, inboundPortsExclude string) corev1.Container {
+	var env []corev1.EnvVar
+	if mode == ProxyInitModeEnforceDrop {
```
mrsabath: suggestion: `if mode == ProxyInitModeEnforceDrop { ... } else { ... }` means any value that isn't exactly `"enforce-drop"` falls through to redirect mode. Today both callsites use typed constants so risk is contained, but `BuildProxyInitContainer` is exported — a future caller passing `"enfore-drop"` (typo) silently disables enforcement, which is the worst possible failure mode for a fail-closed guard.

Two cheap options:

```go
// (a) explicit default branch
switch mode {
case ProxyInitModeEnforceDrop:
    // ...
case ProxyInitModeRedirect:
    // ...
default:
    builderLog.Error(nil, "unknown proxy-init mode", "mode", mode)
    return corev1.Container{}
}

// (b) typed mode
type ProxyInitMode string
const (
    ProxyInitModeRedirect    ProxyInitMode = "redirect"
    ProxyInitModeEnforceDrop ProxyInitMode = "enforce-drop"
)
```

Option (b) lets the compiler catch the typo. A typo in a fail-closed guard turns it into fail-open — 千里之堤，溃于蚁穴 (a thousand-li dike collapses from an ant-hole).

kagenti-operator/internal/webhook/injector/container_builder_test.go
```diff
@@ -282,6 +284,57 @@ func TestBuildProxyInitContainer_WithBothExcludes(t *testing.T) {
...
+		t.Errorf("proxy-sidecar RunAsUser = %d, want Proxy.UID %d", got, cfg.Proxy.UID)
+	}
+	if sc.RunAsGroup == nil || *sc.RunAsGroup != cfg.Proxy.UID {
+		t.Errorf("proxy-sidecar RunAsGroup = %v, want Proxy.UID %d", sc.RunAsGroup, cfg.Proxy.UID)
+	}
+}
+
+// enforce-drop mode emits MODE / PROXY_UID / CLUSTER_CIDRS and none of the
+// redirect-only vars (POD_IP, OUTBOUND_PORTS_EXCLUDE).
+func TestBuildProxyInitContainer_EnforceDrop(t *testing.T) {
```
mrsabath, line 309: suggestion: the new test covers enforce-drop env vars, but there's no test that an invalid mode is rejected. Today `BuildProxyInitContainer("enfore-drop", "", "")` (typo) would silently render as redirect — exactly the failure this PR exists to prevent.

If you keep the string-with-fallthrough approach, add a negative test:

```go
func TestBuildProxyInitContainer_InvalidMode(t *testing.T) {
    builder := newTestBuilder(t)
    c := builder.BuildProxyInitContainer("enfore-drop", "", "")
    // assert: returned zero-value, OR env doesn't contain MODE=enforce-drop
    //         AND doesn't contain redirect-only PROXY_PORT
}
```

Goes away if you adopt the typed-mode approach in the other comment.

kagenti-operator/internal/webhook/config/types.go
```diff
@@ -119,6 +151,31 @@ func (c *PlatformConfig) Validate() error {
...
+	// enforce-drop guard. Validate them at load time so a misconfig fails fast with
+	// a clear message rather than: (a) an empty list silently falling back to the
+	// Kind-shaped 10.0.0.0/8 default in init-iptables.sh, or (b) a malformed entry
+	// crashing the proxy-init container under `set -e` with a cryptic iptables error.
+	if c.Proxy.EgressEnforcement == EgressEnforcementEnforce {
+		if len(c.Proxy.ClusterCIDRs) == 0 {
+			return fmt.Errorf("proxy.clusterCIDRs must be non-empty when proxy.egressEnforcement is %q (set the cluster's pod+service CIDRs)", EgressEnforcementEnforce)
+		}
+		for _, cidr := range c.Proxy.ClusterCIDRs {
+			if _, _, err := net.ParseCIDR(cidr); err != nil {
```
mrsabath, line 177: nit: `net.ParseCIDR(cidr)` per entry catches malformed syntax (good), but doesn't reject overlapping ranges or warn on IPv6 in IPv4-only deployments. Probably intentional — overlapping CIDRs are fine for iptables — but worth a one-line comment so future readers don't assume the validation is exhaustive:

```go
// Syntactic validation only — overlapping ranges and IPv4/IPv6 mixing are
// accepted (iptables handles both); we only reject malformed strings here.
```

mrsabath (approved):
## Summary

Default-off opt-in for proxy-sidecar egress enforcement. The UID fix (1001 → `cfg.Proxy.UID`) is the right cleanup — the prior hardcode had proxy and app sharing a UID, which would have defeated any UID-owner exemption.

Validation in `types.go:122-176` is solid: switch covers empty/off/enforce, requires non-empty `ClusterCIDRs` when enforce, and `net.ParseCIDR`s each entry. Helm `values.yaml` documents the OCP/EKS override requirement clearly. Redirect path is byte-identical to pre-PR (verified), so envoy-sidecar mode is unaffected. Proxy-init's existing `Privileged: false` + `NET_ADMIN, NET_RAW` capabilities posture is unchanged.

Only meaningful gap is that the new public `mode` param has no validation — a typo at any future callsite silently turns the fail-closed guard into fail-open. Worth tightening before this rolls out further; not blocking since today's two callsites both use typed constants.

**Areas reviewed**: Go (webhook config, injector, container builder), tests, Helm values, CI
**Commits**: 2 (both signed-off, DCO ✅)
**CI status**: ✅ all required checks passing (E2E still running at review time)

Assisted-By: Claude Code

kagenti-operator/internal/webhook/injector/container_builder.go
huang195: Adopted the typed-mode approach **and** kept an explicit fail-closed `default`. Worth noting: a named type alone doesn't fully protect — an untyped string literal still implicitly converts to `ProxyInitMode` at a callsite, so `BuildProxyInitContainer("enfore-drop", …)` would still compile. The `default` branch is what actually closes the gap: it logs an error and returns a zero-value container, so a typo'd mode breaks injection visibly instead of degrading to redirect. Fixed in d280093.

kagenti-operator/internal/webhook/injector/container_builder_test.go
huang195: Added `TestBuildProxyInitContainer_InvalidMode` in d280093 — passes `ProxyInitMode("enfore-drop")` and asserts a zero-value container (empty name/image/env), exercising the new fail-closed default branch.

kagenti-operator/internal/webhook/config/types.go
huang195: Added the clarifying comment in d280093 — validation is syntactic only; overlapping ranges and v4/v6 mixing are accepted (iptables and the init script split v4/v6 themselves).

PR #407, opened 2026-06-05
feat: Always-on proxy-sidecar egress enforcement (enforce-redirect), remove flag

huang195:
## Summary

Makes proxy-sidecar egress enforcement **always-on** and **removes the `proxy.egressEnforcement` flag**.

PR #406 gated enforcement behind `egressEnforcement` (default `off`) because the only enforcement available was fail-closed **DROP**, which breaks agents that ignore `HTTP_PROXY`. [kagenti-extensions#485](https://github.com/kagenti/kagenti-extensions/pull/485) adds the missing half — a transparent listener + `MODE=enforce-redirect` that **captures** bypass egress (REDIRECT to the transparent listener, recovering the original destination via `SO_ORIGINAL_DST`) instead of dropping it. Because nothing breaks, enforcement can be on by default with no knob.

## Changes
- **constants**: add `ProxyInitModeEnforceRedirect`; keep `ProxyInitModeEnforceDrop` as a documented fallback the operator no longer emits.
- **container_builder**: enforce-redirect branch emits `MODE` / `PROXY_UID` / `CLUSTER_CIDRS` / `TRANSPARENT_PORT` (new `ProxyConfig.TransparentPort`, default `8082`, matching the authbridge `listener.transparent_proxy_addr` preset).
- **pod_mutator**: the proxy-sidecar / lite branch now **always** injects the enforce-redirect proxy-init (no flag check); envoy-sidecar still uses `redirect`. Idempotent via the existing `containerExists` guard.
- **config**: **remove** `ProxyConfig.EgressEnforcement` (+ its constants, validation, and the Helm value). `ClusterCIDRs` is now validated **unconditionally** (always consumed by the always-on guard); `TransparentPort` gets a port-range check.
- **Helm values**: replace `egressEnforcement` with `transparentPort`; update comments. Rendered platform-defaults ConfigMap verified.

## Tests
- builder emits enforce-redirect env incl. `TRANSPARENT_PORT`;
- proxy-sidecar **always** injects proxy-init (enforce-redirect), idempotent; default-off subtest removed;
- `ClusterCIDRs` always validated; new `TransparentPort` validation test;
- webhook envtest now expects proxy-init in proxy-sidecar defaults.

`go build` / `go vet` / `gofmt` / full `go test ./internal/webhook/...` green. The one `golangci-lint` finding (`pod_mutator.go:187` goconst) is pre-existing and unrelated.

## ⚠️ Do not merge until images ship
Depends on kagenti-extensions#485 being released in the `proxy-init` **and** `authbridge`/`authbridge-lite` images (the init script's `enforce-redirect` mode + the proxy's transparent listener). The install chart owns image versions; this PR does not pin them.

## Preconditions for the always-on guarantee (documented, not enforced here)
- **UID separation** — agents must not run as `Proxy.UID` (1337) nor hold `CAP_SETUID`, else they bypass the `--uid-owner` exemption. Confirm via PSA `restricted`/SCC or an admission check.
- **Name-vs-IP policy** — captured traffic gates on the recovered SNI/Host name (extensions#485), which an agent controls independently of the IP it dials; reliable against a cooperative/misconfigured agent, not a hostile one. Hard enforcement would need IP-set allowlists or SNI/cert cross-checks.

### DIFF
--- a/kagenti-operator/internal/webhook/config/types.go
+++ b/kagenti-operator/internal/webhook/config/types.go
@@ -41,36 +41,26 @@ type ImageConfig struct {
 	PullPolicy corev1.PullPolicy `json:"pullPolicy" yaml:"pullPolicy"`
 }
 
-// EgressEnforcement values for ProxyConfig.EgressEnforcement.
-const (
-	EgressEnforcementOff     = "off"
-	EgressEnforcementEnforce = "enforce"
-)
-
 type ProxyConfig struct {
 	Port             int32 `json:"port" yaml:"port"`
 	UID              int64 `json:"uid" yaml:"uid"`
 	InboundProxyPort int32 `json:"inboundProxyPort" yaml:"inboundProxyPort"`
 	AdminPort        int32 `json:"adminPort" yaml:"adminPort"`
 
-	// EgressEnforcement controls the proxy-sidecar fail-closed egress guard.
-	//   "off" (default): the workload routes egress through the forward proxy
-	//     via HTTP_PROXY only — cooperative and bypassable.
-	//   "enforce": a proxy-init container installs the enforce-drop iptables
-	//     guard, forcing all external egress through the forward proxy regardless
-	//     of whether the workload honors HTTP_PROXY.
-	// envoy-sidecar mode is unaffected — it already enforces egress structurally
-	// via the transparent redirect, so this knob is consulted only on the
-	// proxy-sidecar / lite path. Migrate the default off->enforce in a future
-	// release once ClusterCIDRs sourcing is validated across distros.
-	EgressEnforcement string `json:"egressEnforcement" yaml:"egressEnforcement"`
+	// TransparentPort is the forward proxy's transparent listener port — the
+	// REDIRECT target the enforce-redirect proxy-init guard sends captured
+	// external TCP egress to. It MUST match the authbridge proxy-sidecar
+	// listener.transparent_proxy_addr (default :8082).
+	TransparentPort int32 `json:"transparentPort" yaml:"transparentPort"`
 
 	// ClusterCIDRs are the in-cluster ranges (pods / services / DNS) that the
-	// enforce-drop guard allows direct; everything else egressing the pod is
-	// dropped. The default 10.0.0.0/8 is Kind-shaped (pods 10.244/16 + services
-	// 10.96/16). OCP/EKS MUST override this (e.g. OCP services 172.30.0.0/16,
-	// pods 10.128.0.0/14 — 172.30/16 is outside 10/8) or in-cluster service
-	// traffic will be dropped. Only used when EgressEnforcement == "enforce".
+	// enforce-redirect guard allows direct; external TCP is REDIRECTed to the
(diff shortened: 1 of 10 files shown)


kagenti-operator/internal/webhook/config/types.go
```diff
@@ -41,36 +41,26 @@ type ImageConfig struct {
...
-	// 10.96/16). OCP/EKS MUST override this (e.g. OCP services 172.30.0.0/16,
-	// pods 10.128.0.0/14 — 172.30/16 is outside 10/8) or in-cluster service
-	// traffic will be dropped. Only used when EgressEnforcement == "enforce".
+	// enforce-redirect guard allows direct; external TCP is REDIRECTed to the
+	// transparent listener and external non-TCP is dropped. The default
+	// 10.0.0.0/8 is Kind-shaped (pods 10.244/16 + services 10.96/16). OCP/EKS
+	// MUST override this (e.g. OCP services 172.30.0.0/16, pods 10.128.0.0/14 —
+	// 172.30/16 is outside 10/8) or in-cluster service traffic will be dropped.
+	// Egress enforcement is always-on for proxy-sidecar / lite, so this is
+	// always consumed there; envoy-sidecar (transparent redirect) does not use it.
```
mrsabath, line 63: **suggestion (non-blocking):** `EgressEnforcement` is removed from the values schema, but the config loader uses plain `yaml.Unmarshal` (no `KnownFields(true)`) — so a stale `proxy.egressEnforcement:` left in an operator's existing values override is silently ignored rather than erroring. That's the benign failure mode (enforcement just becomes always-on anyway), so no upgrade-time crash. Still worth a one-liner in the release/upgrade notes so operators aren't surprised the knob no longer does anything.

mrsabath (approved):
Clean refactor — removes the `proxy.egressEnforcement` gate and makes proxy-sidecar/lite egress enforcement always-on, renaming the mode `enforce-drop` → `enforce-redirect` (captures bypass egress via REDIRECT to AuthBridge's transparent listener instead of dropping it).

I verified the cross-repo env-var contract against kagenti-extensions#485 and it matches exactly: `MODE=enforce-redirect`, `TRANSPARENT_PORT` (default `8082`), `PROXY_UID`, and `CLUSTER_CIDRS` are all produced here and consumed by the init script there, and the authbridge `transparent_proxy_addr` default `:8082` lines up with the operator's `TransparentPort` default. The flag removal, de-nested unconditional `ClusterCIDRs` validation, new `TransparentPort` range check, and the test rewrites all look correct.

**Areas reviewed**: Go (config/injector/tests), Helm values, cross-repo contract, commit conventions
**Commits**: 2, both signed-off (DCO ✅)
**CI**: all green (Build, Unit, Integration, E2E, Lint, CodeQL, Trivy, Helm lint, …)
**Dependency gate**: the "do not merge until extensions#485 ships in proxy-init + authbridge images" warning is accurate — extensions#485 is still OPEN, so keep the merge gate.

One non-blocking suggestion inline. LGTM otherwise. 无懈可击 — flawless, not a chink to attack.

PR #383, opened 2026-05-28
feat: create SharedTrust controller for cert-manager cacerts reconciliation

r3v5:
## Summary

- Add event-driven SharedTrust controller that watches cert-manager Certificate objects and reconciles `cacerts`
  secrets for Istio mTLS in `istio-system` and `openshift-ingress`
- On cert data changes, trigger rollout restarts of istiod, istiod-openshift-gateway, and ztunnel so they reload
  the new CA chain
- Verify root CA fingerprints match intermediate CA issuers; delete stale intermediates to force cert-manager
  re-issuance on mismatch
- Controller registers conditionally — skips startup if cert-manager CRDs are absent

## Jira
[RHAIENG-4901](https://redhat.atlassian.net/browse/RHAIENG-4901)

## Next steps
When current PR is merged, I will submit this change `kagenti/scripts/ocp/setup-kagenti.sh` (remove operator-owned steps from _ensure_rhoai_shared_trust; retain cluster-admin/static install steps) 

  ## Test plan

  - [x] Unit tests: fingerprint verification, cacerts assembly, all reconcile paths (fake client)
  - [x] E2E tests: Kind cluster with cert-manager — cacerts creation, rebuild after deletion, restart annotations,
  graceful missing workload handling
  - [x] Manual verification on Kind with Istio ambient mode — full mTLS connectivity confirmed
  - [x] Manual verification on OpenShift (ROSA) with OSSM3 — cacerts created, fingerprints match, istiod loads our
  CA chain, ztunnel mTLS with SPIFFE identities, cert rotation triggers rebuild
  - [x] Run: `go test ./internal/controller/ -run "TestReconcile_|TestVerifyCA|TestBuildCacerts" -v`
  - [x] Run: `export KIND_EXPERIMENTAL_PROVIDER=podman && CONTAINER_TOOL=podman KIND_CLUSTER=kagenti go test
  ./test/e2e/ -v -count=1 -timeout 20m --ginkgo.focus="SharedTrust"`
 - [x] Manual verification on OpenShift 4.19.18 (ROSA) with RHOAI 3.4.0, OSSM3 v3.2.2, Istio v1.27.5

PR #409, opened 2026-06-06
fix(chart): grant cert-manager certificates RBAC to the operator ClusterRole

huang195:
## Problem

On a fresh chart install, the operator (`kagenti-controller-manager`) **crashloops** and never becomes Ready, which silently breaks agent onboarding: agent pods hang in `Init` with

```
MountVolume.SetUp failed for volume "kagenti-keycloak-client-credentials-<hash>": secret not found
```

## Root cause

The Helm chart's **hand-maintained** ClusterRole (`charts/kagenti-operator/templates/rbac/role.yaml`) drifted from the kubebuilder-generated `config/rbac/role.yaml`: it is **missing the `cert-manager.io` `certificates`/`clusterissuers` `get;list;watch` rules**.

[#383](https://github.com/kagenti/kagenti-operator/pull/383) (SharedTrust controller) added `Watches(&cmv1.Certificate{})` plus the `+kubebuilder:rbac:groups=cert-manager.io,resources=certificates` marker and regenerated `config/rbac/role.yaml` — but did **not** update the chart's separate ClusterRole. Since installs render RBAC from the chart, the deployed ServiceAccount `kagenti-system:controller-manager` cannot list/watch cert-manager `Certificates`:

```
failed to list *v1.Certificate: certificates.cert-manager.io is forbidden:
User "system:serviceaccount:kagenti-system:controller-manager" cannot list resource "certificates" ...
```

That informer never syncs → the manager misses its cache-sync deadline and **exits 1 (~2.5 min crashloop)** → **no controllers run**, including `ClientRegistrationReconciler`. So agents are never registered in Keycloak, their per-agent client-credentials secrets are never created, and the webhook-mounted secret volume never resolves → permanent `FailedMount`.

Reproduced on `weather-agent` (team1) on a fresh install of `main`; matches a user report (#383-era). Confirmed the fix on a live cluster: granting these rules + restarting the operator → Ready → reconciler creates the secret → agent reaches `2/2 Running`.

## Fix

Add the `cert-manager.io` `certificates`/`clusterissuers` `get;list;watch` rules to the chart ClusterRole to match the generated `config/rbac/role.yaml`. `helm lint` + `helm template` verified.

## Follow-up (separate)

The chart ClusterRole is maintained **independently** of `config/rbac/role.yaml` with no sync step, so any future `+kubebuilder:rbac` marker can silently drift again (this is the second controller bitten — see also the MLflow/OTEL RBAC edits in #404/#378 that hand-patch the same file). Worth generating the chart role from the kubebuilder role, or adding a CI check that fails on drift.

### DIFF
--- a/charts/kagenti-operator/templates/rbac/role.yaml
+++ b/charts/kagenti-operator/templates/rbac/role.yaml
@@ -63,6 +63,15 @@ rules:
   - get
   - update
 - apiGroups:
+  - cert-manager.io
+  resources:
+  - certificates
+  - clusterissuers
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - agent.kagenti.dev
   resources:
   - agentcards


mrsabath (approved):
Verified the fix against the kubebuilder-generated `kagenti-operator/config/rbac/role.yaml` on `main`: the added `cert-manager.io` `certificates`/`clusterissuers` `get;list;watch` block matches the generated role byte-for-byte, and the `+kubebuilder:rbac` markers in `sharedtrust_controller.go` (#383) confirm the root cause. The chart ClusterRole on `main` was indeed missing these rules. Read-only verbs only — no over-grant. All CI green incl. E2E.

Agreed the chart-role-vs-generated-role drift deserves a CI guard as a follow-up. 亡羊补牢 (mend the pen after losing a sheep) — better to close this gap before a third controller gets bitten.

**Areas reviewed**: Helm/RBAC, commit/PR conventions, security, CI · **Commits**: 1, signed-off: yes · **CI**: passing

PR #411, opened 2026-06-08
feat: patch kagenti-ui-config ConfigMap with MLflow dashboard URL at startup

ChristianZaccaria:
## Summary
Adds a startup runnable (gated by `--enable-mlflow`) that patches the `kagenti-ui-config` ConfigMap with the MLflow dashboard URL derived from the MLflow CR's `status.url`. This removes the need for `setup-kagenti.sh` to handle UI config patching after MLflow Route creation.

## Changes
- `internal/bootstrap/uiconfig.go` — `UIConfigBootstrapRunnable` that discovers MLflow CR and patches the ConfigMap once at startup
- `internal/bootstrap/uiconfig_test.go` — unit tests (idempotency, skip when no MLflow/ConfigMap)
- `cmd/main.go` — register the runnable under `--enable-mlflow`
- `Makefile` — fix `docker-buildx` target (Dockerfile already has `--platform`)

## Test Plan
- [x] Unit tests pass (`go test ./internal/bootstrap/ -run TestUIConfig`)
- [x] Verified on OCP cluster with RHOAI MLflow: ConfigMap patched correctly
- [ ] CI green

### DIFF
--- a/kagenti-operator/internal/bootstrap/uiconfig.go
+++ b/kagenti-operator/internal/bootstrap/uiconfig.go
@@ -1,0 +1,105 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package bootstrap
+
+import (
+	"context"
+	"strings"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kagenti/operator/internal/mlflow"
+)
+
+const (
+	uiConfigMapName       = "kagenti-ui-config"
+	mlflowDashboardURLKey = "MLFLOW_DASHBOARD_URL"
+)
+
+// UIConfigBootstrapRunnable patches the kagenti-ui-config ConfigMap with the
+// MLflow dashboard URL. Runs once at startup; skips gracefully if MLflow or
+// the UI ConfigMap are not present.
+type UIConfigBootstrapRunnable struct {
+	Client    client.Client
+	APIReader client.Reader
+	Namespace string
+	Log       logr.Logger
+}
+
+func (r *UIConfigBootstrapRunnable) Start(ctx context.Context) error {
+	log := r.Log.WithName("ui-config-bootstrap")
+
+	dashboardURL := r.discoverDashboardURL(ctx, log)
+	if dashboardURL == "" {
+		return nil
+	}
+
+	cm := &corev1.ConfigMap{}
+	key := types.NamespacedName{Name: uiConfigMapName, Namespace: r.Namespace}
+	if err := r.APIReader.Get(ctx, key, cm); err != nil {
+		if errors.IsNotFound(err) {
+			log.V(1).Info("UI ConfigMap not found, skipping (UI likely not deployed)")
+			return nil
+		}
+		log.Error(err, "Failed to read UI ConfigMap")
+		return nil
+	}
+
+	if cm.Data != nil && cm.Data[mlflowDashboardURLKey] == dashboardURL {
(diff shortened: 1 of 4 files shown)


kagenti-operator/internal/bootstrap/uiconfig.go
```diff
@@ -0,0 +1,105 @@
...
+	APIReader client.Reader
+	Namespace string
+	Log       logr.Logger
+}
+
+func (r *UIConfigBootstrapRunnable) Start(ctx context.Context) error {
+	log := r.Log.WithName("ui-config-bootstrap")
+
+	dashboardURL := r.discoverDashboardURL(ctx, log)
+	if dashboardURL == "" {
```
cwiklik, line 52: **suggestion:** One-shot bootstrap silently gives up if no MLflow CR is `Available` yet. `Start()` runs once; on a fresh install the MLflow Route is typically admitted *after* the operator boots, so `discoverDashboardURL` returns `""`, this returns `nil`, and the ConfigMap is never patched until the operator pod restarts. This couples correct behavior to startup ordering. Consider a watch/requeue on MLflow status (or a periodic retry) so the patch is applied whenever MLflow becomes ready — the operator handling ordering rather than depending on it.

kagenti-operator/internal/bootstrap/uiconfig.go
```diff
@@ -0,0 +1,105 @@
...
+	if cm.Data != nil && cm.Data[mlflowDashboardURLKey] == dashboardURL {
+		log.V(1).Info("MLflow dashboard URL already set")
+		return nil
+	}
+
+	if cm.Data == nil {
+		cm.Data = make(map[string]string)
+	}
+	cm.Data[mlflowDashboardURLKey] = dashboardURL
+	if err := r.Client.Update(ctx, cm); err != nil {
```
cwiklik, line 76: **suggestion:** The PR title says "patch" but this is a read-modify-write `Update`. If another writer (e.g. the UI deploy) updates the CM between `Get` and `Update`, you get a resourceVersion conflict → logged and dropped (`return nil`), patch lost. A `client.Patch` (merge/strategic) of the single `MLFLOW_DASHBOARD_URL` key avoids the conflict window and matches the stated intent.

kagenti-operator/internal/bootstrap/uiconfig.go
```diff
@@ -0,0 +1,105 @@
...
+	return nil
+}
+
+func (r *UIConfigBootstrapRunnable) NeedLeaderElection() bool {
+	return true
+}
+
+func (r *UIConfigBootstrapRunnable) discoverDashboardURL(ctx context.Context, log logr.Logger) string {
+	list := &mlflow.MLflowList{}
+	if err := r.APIReader.List(ctx, list); err != nil {
```
cwiklik, line 91: **nit:** `List` is cluster-wide and returns the first `Available` CR by list order — non-deterministic if multiple MLflow CRs exist across namespaces. If a single instance is expected, scope with `client.InNamespace(r.Namespace)` or document the single-instance assumption.

cwiklik (approved):
Small, well-scoped feature moving UI-config patching from `setup-kagenti.sh` into the operator. Clean code, correctly gated under `--enable-mlflow`, uses `APIReader` for direct reads, and `NeedLeaderElection() = true` to avoid multi-replica races. Strong test coverage (patch, idempotency, both skip paths, trailing-slash normalization). No must-fix issues.

Two suggestions are about robustness, not correctness — the one-shot-at-startup model (#1) is the most substantive: it works once MLflow is ready but won't self-heal if MLflow lags operator startup.

Verified the Makefile `docker-buildx` change is correct: `kagenti-operator/Dockerfile` already has `FROM --platform=$BUILDPLATFORM`, so removing the `Dockerfile.cross` sed hack is safe.

Nit: commit subject uses `feat:` without the emoji prefix Kagenti's commit convention favors (`:sparkles: feat:`) — style only, DCO is green. All 16 CI checks passing.

kagenti-operator/internal/bootstrap/uiconfig.go
ChristianZaccaria: Documented the single-instance assumption — Kagenti deploys one MLflow per platform, so cluster-wide list is intentional and the first Available CR is always the correct one.

kagenti-operator/internal/bootstrap/uiconfig.go
ChristianZaccaria: Switched to `client.MergeFrom` patch — now only the `MLFLOW_DASHBOARD_URL` key is patched without touching the rest of the ConfigMap, avoiding resourceVersion conflicts with concurrent writers.

kagenti-operator/internal/bootstrap/uiconfig.go
ChristianZaccaria: Thanks for the suggestion, I added a retry loop with backoff (~4 min window), previously the kagenti setup script would wait 2mins. A full controller/watch felt heavyweight for a one-shot ConfigMap patch, the retry covers the typical MLflow startup lag. If MLflow takes significantly longer, a pod restart or manual reconciliation could handle this.

PR #415, opened 2026-06-09
Feat:  Bundle Service serving OPA authorization policy bundles

### LINKED ISSUE #468: feature: OPA Bundle Service
### Feature Description

OPA Bunder Service see Epic: https://github.com/kagenti/kagenti/issues/1789

## OPA Bundle Service 
The OPA Bundle Service implements a standard OPA service for distributing OPA bundles. The bundles are distributed based on client id (SPIFFE). The service generates a dedicated bundle for each agent and each tool. Per‑agent and per-tool bundles enforce strict isolation: each agent receives only the policies relevant to its own identity, reducing blast radius and preventing agents from inferring or accessing rules intended for others.
Each bundle is tied to the component’s SPIFFE identity, enabling Security Engineers and SREs to define four separate Rego files per agent or tool:

- Inbound request policy
- Inbound response policy
- Outbound request policy
- Outbound response policy

Inbound policies evaluate what is allowed to reach the agent or tool, while outbound policies evaluate what the agent or tool itself is allowed to call or return.

This design is:

- IdP-agnostic — Works with any identity provider (Keycloak, Entra ID, Okta, custom IdPs). Policies reference identity claims, not a specific token system.
- Runtime context — Decisions are made at request time with full context.
- Kubernetes-native — Policies are managed as CRs with schema validation, RBAC, and GitOps workflows — aligned with Kubernetes access-control practices.
- Expressive — Rego supports RBAC, ABAC, relationship-based rules, content inspection, and combinational logic in a single policy language.
- No vendor lock-in — OPA is CNCF-graduated. Switching or adding identity providers requires no policy rewrites.

### Want to contribute?

- [x] I would like to work on this issue.

### Additional Context

_No response_

davidhadas:
Introduce the bundle-service component, an HTTP server that assembles and serves OPA-compatible authorization bundles to AuthBridge sidecar clients. Each bundle is tailored to the requesting client's identity and built from three tiers of AuthorizationPolicy CRs: global, namespace, and client-scoped.

Key design decisions:
- Decision logic lives in the global AuthorizationPolicy CR, not in code, giving platform engineers full control over tier combination semantics (override support, tier removal, AND/OR changes)
- Bounded concurrency (semaphore + singleflight) protects the Kubernetes API server and etcd from thundering herd on cluster restart
- Three-layer caching (ETag, bundle, policy) minimizes rebuild pressure
- ETag-based conditional responses (304 Not Modified) reduce bandwidth

Includes:
- AuthorizationPolicy CRD types and conversion helpers
- Bundle builder (tar.gz with OPA manifest)
- Multi-tier policy cache with scope-aware invalidation
- Kubernetes informer-based watcher with scope indexing
- SPIFFE ID identity parsing
- Deployment manifests, RBAC, NetworkPolicy, default global policy CR
- Kind development script (hack/bundle-service-kind.sh)
- Architecture and operational documentation
- Comprehensive unit tests for all packages

## Related issue(s)
[#1789](https://github.com/kagenti/kagenti/issues/1789)

#fixes: https://github.com/kagenti/kagenti-operator/issues/468


Alan-Cha (approved):
## Summary

Strong implementation of the bundle-service component with comprehensive test coverage, good architecture documentation, and proper security considerations. The three-tier caching strategy and bounded concurrency design show careful attention to production readiness.

**Areas reviewed:** Go code (3400+ lines), Kubernetes manifests, CRDs, deployment configs, documentation, shell scripts, Dockerfile
**Commits:** 1 commit, signed-off: ✅ yes
**CI status:** 14/15 passing (E2E tests failing)

---

## Required before merge

### 1. E2E Test Failure (must-fix)
The E2E Tests check is failing after 25m50s runtime. This needs investigation and resolution before merge. Check the test logs at: https://github.com/kagenti/kagenti-operator/actions/runs/27195746396/job/80286789432

### 2. PR Title Convention (must-fix)
Add emoji prefix per repo conventions. Current: `Feat: Bundle Service...`
Suggested: `✨ Feat: Bundle Service serving OPA authorization policy bundles`

---

## Suggestions (non-blocking)

### Code Quality
- **`cmd/bundle-service/main.go`**: Consider adding graceful shutdown handler to drain in-flight requests during pod restarts (prevents 502s)
- **`internal/bundleservice/handler/handler.go`**: ETag validation logic could be extracted to a helper method for better testability
- Consider adding observability metrics for cache hit/miss rates (singleflight pattern is excellent, metrics would make it visible)

### Documentation
- **`README.md`**: Add direct link to `hack/bundle-service-kind.sh` in the bundle service section
- **`ARCHITECTURE.md`**: Excellent content - consider adding troubleshooting section for common operational issues
- Consider documenting the ETag computation strategy in ARCHITECTURE.md (currently only in code comments)

### Testing
- Consider adding integration test coverage for multi-tier policy combination logic (unit tests are thorough, integration would validate end-to-end flow)

### Security
- NetworkPolicy: ✅ properly restricts ingress to port 8080
- RBAC: ✅ appropriately scoped to read-only AuthorizationPolicies
- Consider documenting threat model for bundle tampering (likely covered by SPIFFE mTLS but worth explicit mention)

---

## What I particularly like

1. **Design decision to keep authorization logic in CRs** - This gives platform engineers full control without code changes. Brilliant.
2. **Bounded concurrency with singleflight** - Protects K8s API server from thundering herd. Production-ready thinking.
3. **Three-layer caching** - ETag → bundle → policy tiers show deep understanding of performance optimization
4. **Comprehensive test coverage** - Every package has tests, conversion helpers are well-tested
5. **Documentation quality** - ARCHITECTURE.md explains not just *what* but *why*

Excellent work overall. Fix the E2E tests and title, and this is ready to merge.

davidhadas:
@Alan-Cha 
Thank you for the review
1. E2e tests failing are not related to the bundle service. Not sure why they occur  - maybe prior issue?
2. Adding EMoji to title leads to failure of PR Verifier

kagenti-operator/internal/bundleservice/identity/identity.go
```diff
@@ -0,0 +1,61 @@
...
+// Verifier verifies that the caller is authorized to request bundles for the
+// given identity. Implementations may check mTLS client certificates (SPIFFE)
+// or JWT bearer tokens (Kubernetes service accounts).
+type Verifier interface {
+	Verify(r *http.Request, id ClientIdentity) error
+}
+
+// NoopVerifier performs no verification. Used when the service does not yet
+// terminate TLS or validate tokens (verification is handled upstream).
+type NoopVerifier struct{}
```
mrsabath, line 57: Tracking note (not a blocker): identity is parsed from the client-controlled `?spiffe=` query param and `NoopVerifier` performs no checks, so on its own the service trusts whatever a caller claims. `ARCHITECTURE.md` documents this as intentional — verification is deferred upstream, the NetworkPolicy restricts ingress to `kagenti.dev/authbridge`-labeled pods, and this `Verifier` interface is the seam for the real check. That's a reasonable MVP trust boundary. The ask is just to land an mTLS-SAN or JWT verifier here **before** the service is reachable beyond the AuthBridge sidecar mesh, since at that point the `?spiffe=` param becomes a cross-client bundle-fetch vector. Worth a tracking issue linked off #1789.

kagenti-operator/config/bundleservice/networkpolicy.yaml
```diff
@@ -0,0 +1,45 @@
...
+              k8s-app: kube-dns
+      ports:
+        - protocol: UDP
+          port: 53
+        - protocol: TCP
+          port: 53
+    # Allow access to Kubernetes API server
+    - to:
+        - ipBlock:
+            cidr: 0.0.0.0/0
```
mrsabath, line 40: suggestion: egress is `0.0.0.0/0` on 443/6443. The DNS rule above is nicely scoped; for the API-server rule, consider narrowing to the API server's ClusterIP / service CIDR where the environment makes that practical, so the policy actually constrains egress. Fine to leave for a follow-up if the CIDR isn't stable across target clusters.

mrsabath (commented):
Reviewed the full bundle-service. Really solid foundation — clean package separation, genuinely comprehensive tests, distroless/non-root image, least-privilege RBAC, and the concurrency model (per-client singleflight key, semaphore released on all paths via `defer`, RWMutex-guarded caches, clean informer lifecycle) all check out. Per-client bundle scoping is correct — no cross-tenant policy leakage — and the tar builder guards against path traversal.

**On CI:** E2E is red, but the only failing spec is `Skill Discovery › should populate linkedSkills from annotation` (`e2e_test.go:2161`) — unrelated to this PR (it touches no skill-discovery code), and `main` is green on the same suite. Looks like a flake; worth a re-run to get it green before merge rather than merging red.

A few non-blocking notes inline. The one I'd most want tracked: identity today comes from the client-controlled `?spiffe=` query parameter with `NoopVerifier`, which `ARCHITECTURE.md` documents as intentional (verification deferred upstream, the `Verifier` interface as the seam, NetworkPolicy limiting ingress to AuthBridge sidecars). That's fine for the MVP trust boundary, but the mTLS/JWT verifier should land before this is reachable beyond the sidecar mesh — might be worth a tracking issue.

Holding my own approval until E2E is green; @Alan-Cha has already approved.

**Areas reviewed**: Go (identity/handler/provider/cache/watcher/bundle), security/authz, concurrency, RBAC/NetworkPolicy, Dockerfile, CI · **Commits**: 1, signed-off: yes · **CI**: E2E failing (unrelated flake — Skill Discovery)

kagenti-operator/cmd/bundle-service/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+COPY go.sum go.sum
+RUN go mod download
+
+COPY cmd/bundle-service/ cmd/bundle-service/
+COPY api/ api/
+COPY internal/bundleservice/ internal/bundleservice/
+
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -o bundle-service ./cmd/bundle-service/
+
+FROM gcr.io/distroless/static:nonroot
```
mrsabath, line 20: nit: `gcr.io/distroless/static:nonroot` isn't pinned by digest. Pinning to `...@sha256:<digest>` makes the build reproducible and satisfies Scorecard's pinned-dependencies check. (Consistent with how the chart-RBAC/Scorecard items have been handled elsewhere — fine to defer.)

kagenti-operator/internal/bundleservice/watcher/watcher.go
```diff
@@ -0,0 +1,184 @@
...
+		slog.Info("global policy changed, invalidating all caches")
+		w.policyCache.InvalidateGlobal()
+		w.etagCache.InvalidateAll()
+		w.bundleCache.InvalidateAll()
+
+	case "namespace":
+		ns := u.GetNamespace()
+		slog.Info("namespace policy changed", "namespace", ns)
+		w.policyCache.InvalidateNamespace(ns)
+		w.etagCache.InvalidateAll()
```
mrsabath, line 173: nit (perf, not correctness): on a `namespace`-tier policy change this calls `etagCache.InvalidateAll()` / `bundleCache.InvalidateAll()`, which also evicts clients in *other* namespaces. Bundles stay correct (the `policyCache.InvalidateNamespace(ns)` above is properly scoped), so this is just extra cache churn / 200s where a 304 would do. An `InvalidateFunc` keyed on `strings.HasPrefix(agentID, ns+"/")` would scope it. Low priority.

davidhadas:
@mrsabath  - 
Created  https://github.com/kagenti/kagenti/issues/1860
A second attempt of e2e tests led to FAIL! -- 35 Passed | 1 Failed | 0 Pending | 5 Skipped
Better than before - so instability issue - I will try again

pdettori:
@davidhadas CI has been fixed - please rebase from upstream main and force push so we can check now with full CI

davidhadas:
**What should we use as transport?  Bundle Service (HTTP) or ConfigMap Distribution**

The bundle service current design implemented in this PR uses the standard OPA bundle protocol (HTTP polling) to distribute policies to AuthBridge sidecars. @rhuss proposed that the operator controls the sidecar pod spec, the bundle service will push rendered policies via ConfigMaps instead, eliminating the HTTP layer entirely.
Propagation delay, Consistency properties both seems similar. The 1MB ConfigMap limit is unlikely to be a practical constraint for Rego policies.

 **Main arguments for the alternative design - ConfigMap transport (the K8s native way)**
- Eliminates the HTTP serving layer: handler, ETag/304, singleflight, three-tier cache, concurrency control — infrastructure that exists to serve HTTP polling efficiently. ConfigMap writes don't need any of it.
- Alow us to use K8s-native mechanisms and integrate the functionality to become a component of the operator (operator already watches AgentRuntime CRDs and controls the sidecar in other ways): removes a separate Deployment, Service, ServiceAccount, RBAC, and NetworkPolicy.

**Arguments for the current design -  HTTP transport (the OPA native way)**
- OPA SDK embedded in the sidecar support HTTP natively. ConfigMap approach requires designing a custom filesystem-loading to the embedded  OPA — custom code replacing community-maintained code.
- HTTP is the standard OPA bundle transport mechanism: Keeps the system open — customers can point AuthBridge at any web server to obtain bundles.  
- An HTTP Bundle service we design can serve bundles to Authorino and potentially future OPA PDPs which become part of the agent platform – this gives us a uniform and consolidated place to build and distribute Access Control rules across the agent platform.

rhuss:
Thanks for the summary David. A few points from the thread that I think should be reflected for a balanced picture:

**Missing from the ConfigMap section:**
- Code reduction: the HTTP serving layer (handler, ETag/304, singleflight, concurrency control, SPIFFE parsing, three-layer cache) is ~2,500 lines that goes away entirely. Total custom code drops from ~4,000 to ~1,500 lines.
- AuthBridge already hot-reloads config from ConfigMaps (confirmed by the AuthBridge team in the thread). This is a proven mechanism, not a hypothetical.
- No availability dependency: ConfigMaps are in etcd (already required). The bundle service must be running and reachable for every poll cycle.

**On "custom code replacing community-maintained code":**
This framing is one-sided. The bundle service itself is 4,000 lines of custom code (HTTP handler, caching, singleflight, identity parsing, bundle compilation). The ConfigMap approach reduces total custom code. The client-side adaptation (filesystem loader or rego.New with file watching) replaces community HTTP client code, yes, but the net custom code is significantly less.

**On "can serve bundles to Authorino and potentially future OPA PDPs":**
This was discussed at length in the thread. Today: Authorino doesn't use OPA bundles, OpenShell doesn't fetch bundles, Keycloak has no OPA support. Building for adoption that hasn't materialized is premature. Additionally, the bundle service is a separate deployment that ships as part of the agent platform. Any external component that depends on it (like Authorino) would require the agent platform to be installed, even when the use case has nothing to do with agents. If this service is meant to be cross-product, it should live as an independent project with its own lifecycle and release process. That packaging perspective should be part of the tradeoff discussion.

**Suggested addition to the summary:**
A third section covering the operational and packaging dimension would make this complete: deployment footprint, ownership boundaries, and whether the agent platform repo is the right home for a distribution service that aspires to serve components beyond AuthBridge.

davidhadas:
@rhuss Thank you for the additions. To clarify — the summary above was generated by Claude with the goal of being neutral. I asked it to focus on points that hadn't already been addressed in our earlier discussion, keeping a concise view of the alternatives.

I think we may be covering the same ground again — I believe most of these points came up previously and were discussed on Slack.

On Authorino specifically: I don't see a dependency between Authorino and the agent platform. What I do see is an opportunity for a centralized authorization service when Authorino is deployed as part of the agent platform — a point I covered in our earlier Slack thread.

It's also worth noting that even if upstream starts with HTTP transport to explore the benefits it may bring, that doesn't commit downstream to adopting it — this isn't a one-way door. We can continue to learn and adapt the transport in upstream as we go, based on what we observe in practice.

Happy to take this to a quick sync if you feel the written thread isn't converging — sometimes a short call is more productive.

davidhadas:
Here is a diagram to help highlight the deployment options under HTTP and the opportunity for openness it offers.

image

pdettori (changes requested):
Requesting changes to resolve the **HTTP bundle service vs ConfigMap distribution** transport decision (the @rhuss thread) before merging. It's the only substantive open question left — CI is green apart from the title check, and the reviews are positive — but it has an outsized impact on what we'd be committing: the two designs differ by ~2,500 lines and by whether this ships as a separate Deployment/Service/RBAC or folds into the operator. Merging now would lock in a contested architecture while the thread is still open. A short sync to land an explicit decision (even "HTTP as an upstream experiment") seems like the fastest path.

_Assisted-By: Claude Code_

davidhadas:
@pdettori, @mrsabath , @rhuss 
I suggest we merge this  - it is a suitable first design for Upstream and it is OPA compliant. 

It is understood as pointed by @rhuss that downstream is likely to choose a more K8s native solution using ConfigMaps and that from the downstream perspective there may be less value at this point to remain OPA compliant. Still, I think it is a good design choice for upstream and will allow continuing the efforts to embed end to end  authorization services and explore the value it offers. 

Merging this does not prevent upstream to later switch to a  K8s native solution or add such an option if so desired.

pdettori (approved):
LGTM

rhuss:
Any ideas how this could co-exist with a ConfigMap only solution, without using the OPA service ? It's not 'very likely' but downstream won't add this bundle service for sure for the reasons that have been laid out in some extense.

rhuss:
Also any chance to have an alternative way than a separate CR to define the policies, like as part of `AgentRuntime` ? We also don't want to include the `AuthorizationPolicy` for the reasons mentioned. 

I would have love to have some more open discussions around this decision, I'm a bit unsure how to continue mid/downstream if we have to soft-fork that often.

davidhadas:
I suggest we separate the bundle generation logic (how it's constructed from CRDs) from the transport mechanism itself. We can add support for a ConfigMap-based transport as discussed, and then reuse the core package to build the bundles regardless of which transport mechanism is selected.

We could potentially remove the HTTP transport once the ConfigMap approach is in place, though I am not certain that is the right move yet. While dropping the OPA-compliant HTTP option is technically straightforward here, doing so highlights a larger conversation we need to have.

Before we make a final decision on dropping features, we should establish clear alignment on how we balance innovation work and product work—ideally within a single upstream repository. The OPA bundle is a relatively simple case, but we will undoubtedly face more complex features in the future that we want upstream, even if they don't immediately cross the current downstream bar or aren't considered for downstream at present. Let's find a repeatable pattern for handling these scenarios.

rhuss:
@davidhadas Thanks David. Separating bundle generation from transport is the right approach. The policy reconciler logic (CRD watching, tier merging, bundle assembly) is valuable and carries forward regardless of transport mechanism.

For the product: ConfigMap-based transport. The operator writes a ConfigMap, kubelet mounts it, the sidecar reads the file. No new components.

For upstream: the HTTP transport stays as an option for teams that want OPA protocol compatibility. Both share the same bundle generation core.

On the broader question about balancing innovation and product work in a single repo: agreed, we need a repeatable pattern. We started that conversation yesterday and I think the sandbox/graduation model (research org for experimental work, main org for production-ready upstream) gives us a clean boundary. Let's continue working this out with @pdettori and @mrsabath . 

I'm happy to collaborate on the ConfigMap transport implementation, but would be very grateful if you could pick that up and drive it. The core bundle generation code you wrote is the foundation for both paths.

PR #421, opened 2026-06-10
Fix: add TLS connectivity probe to webhook readiness checks

### LINKED ISSUE #420: fix(e2e): Skill Discovery test fails due to insufficient webhook readiness check after controller restart
## Summary

The "Skill Discovery E2E > Feature gate enabled > should populate linkedSkills from annotation" test fails consistently in CI since PR #396 was merged (June 8, 2026). This blocks all PRs — the E2E Tests job is the only failing check across all open PRs.

**Affected**: Every CI run on every branch since June 8.
**Impact**: 1 test fails, 5 skip (cascading), 35 pass.

## Root Cause

**Insufficient webhook readiness verification after controller restart.**

The test checks that the webhook `Endpoints` object has an IP address, but does NOT verify the webhook's TLS server is actually accepting connections. There's a race window (~5-30s) where the endpoint IP is populated but the webhook server hasn't completed TLS setup.

### Failure Chain

1. `EnableSkillDiscovery()` patches the controller deployment with a volume mount → triggers pod restart
2. `WaitForRollout` succeeds (deployment reports Available)
3. Endpoint IP check passes immediately (pod IP is in Endpoints object)
4. Test creates AgentRuntime → controller reconciles → adds `kagenti.io/type` label + config-hash to PodTemplateSpec → triggers deployment rollout
5. ReplicaSet tries to create new pods → Kubernetes calls the `inject.kagenti.io` mutating webhook
6. **Webhook TLS server is not yet ready** → `dial tcp 10.96.50.216:443: connect: connection refused`
7. Pods never get created → deployment never rolls → `linkedSkills` never populates → test times out at 180s

### Why It Started Failing with PR #396

PR #396 changed the "Combined" test (runs immediately before "Skill Discovery") to trigger a rolling update on AR deletion (verifying pods have no sidecars). This increases webhook activity right before the Skill Discovery block tears down and redeploys the controller, making the timing race more likely to manifest.

### Evidence

From CI run `27247825117` (main branch, June 10):
```
Line 506: rollout status completes at 02:03:22
Line 510: endpoint IP check passes at 02:03:22 (same second!)
Line 512: AgentRuntime created at 02:03:22
Line 719: FailedCreate event: "failed calling webhook inject.kagenti.io: 
           dial tcp 10.96.50.216:443: connect: connection refused"
           (persists for 4+ minutes)
```

The endpoint IP exists but the TLS server isn't listening yet.

### Location

- `kagenti-operator/test/e2e/e2e_test.go` — lines 2128-2136 (the `"waiting for webhook endpoint to be ready after restart"` block in the `"Feature gate enabled"` Context's `BeforeAll`)
- `kagenti-operator/test/utils/utils.go` — add a reusable `WaitForWebhookReady()` helper

### Approach

Replace the endpoint-IP-only check with an active TLS connectivity probe. The webhook exposes a health endpoint at `/healthz` on port 9443 inside the pod. The check should:

1. Keep the existing endpoint IP check (fast pre-condition)
2. Add a follow-up that actually connects to the webhook service and verifies TLS handshake succeeds

### Implementation

**In `test/utils/utils.go`**, add a `WaitForWebhookReady` function:

```go
// WaitForWebhookReady waits until the webhook service is actually accepting
// TLS connections, not just until the Endpoints object has an IP.
func WaitForWebhookReady(namespace string, timeout time.Duration) error {
    // Phase 1: endpoint IP exists (fast gate)
    err := wait.PollUntilContextTimeout(context.Background(), 2*time.Second, timeout, true,
        func(ctx context.Context) (bool, error) {
            cmd := exec.Command("kubectl", "get", "endpoints",
                "kagenti-operator-webhook-service", "-n", namespace,
                "-o", "jsonpath={.subsets[0].addresses[0].ip}")
            output, err := Run(cmd)
            if err != nil || output == "" {
                return false, nil
            }
            return true, nil
        })
    if err != nil {
        return fmt.Errorf("webhook endpoint IP not ready: %w", err)
    }

    // Phase 2: TLS port is accepting connections
    return wait.PollUntilContextTimeout(context.Background(), 2*time.Second, timeout, true,
        func(ctx context.Context) (bool, error) {
            // Use kubectl to port-forward and check, or use a Job/exec approach.
            // Simplest: create a temporary pod that curls the webhook, or use
            // kubectl run with --rm to test connectivity.
            cmd := exec.Command("kubectl", "run", "webhook-probe", "--rm", "-i",
                "--restart=Never", "--image=curlimages/curl:latest",
                "-n", namespace, "--",
                "curl", "-sk", "-o", "/dev/null", "-w", "%{http_code}",
                "https://kagenti-operator-webhook-service."+namespace+".svc:443/healthz")
            output, err := Run(cmd)
            if err != nil {
                return false, nil
            }
            // Any HTTP response (even 404) means TLS is up
            return output != "" && output != "000", nil
        })
}
```

**Alternative simpler approach** (no extra pod needed): Use a raw TCP check via a shell command that attempts a TLS connection within the cluster network. Since the e2e tests already have the controller pod running, we can exec into it:

```go
// WaitForWebhookReady verifies the webhook TLS server is accepting connections.
func WaitForWebhookReady(namespace, deployment string, timeout time.Duration) error {
    By("waiting for webhook endpoint IP")
    if err := waitForEndpointIP(namespace, timeout); err != nil {
        return err
    }

    By("waiting for webhook TLS server to accept connections")
    return Eventually(func(g Gomega) {
        // The manager process serves the webhook on :9443 inside the pod.
        // A successful wget --spider proves TLS is up.
        cmd := exec.Command("kubectl", "exec",
            "deploy/"+deployment, "-n", namespace, "--",
            "wget", "--spider", "--no-check-certificate", "-q", "-T", "2",
            "https://localhost:9443/healthz")
        _, err := Run(cmd)
        g.Expect(err).NotTo(HaveOccurred())
    }, timeout, 2*time.Second).Should(Succeed())
}
```

> **Note:** The manager container image is distroless and may not have `wget`/`curl`. If so, use a Go-based TCP dial from a temporary pod, or simply add a short sleep (10s) after the endpoint check as a pragmatic fix. The most robust approach is to create a small utility binary or use the existing `curl` image approach.

**Pragmatic minimal fix** (if the above is too complex): Simply increase the wait and add a fixed delay:

```go
By("waiting for webhook endpoint to be ready after restart")
Eventually(func(g Gomega) {
    cmd := exec.Command("kubectl", "get", "endpoints",
        "kagenti-operator-webhook-service", "-n", controllerNamespace,
        "-o", "jsonpath={.subsets[0].addresses[0].ip}")
    output, err := utils.Run(cmd)
    g.Expect(err).NotTo(HaveOccurred())
    g.Expect(output).NotTo(BeEmpty(), "webhook endpoint not yet populated")
}, 2*time.Minute, 2*time.Second).Should(Succeed())

// Allow webhook TLS server time to start accepting connections
// after the pod is Running and endpoint is populated.
time.Sleep(10 * time.Second)
```

### Where to Apply

This fix needs to be applied at **every** location where `"waiting for webhook endpoint to be ready"` appears after a controller redeploy or restart. In the current `e2e_test.go` these are:
- Line 194 (initial BeforeSuite)
- Line 337 (AuthBridge)
- Line 689 (AgentCard)
- Line 1032 (AgentRuntime)
- Line 1506 (Combined)
- Line 1967 (Skill Discovery BeforeAll)
- **Line 2128** (Skill Discovery Feature gate enabled — the failing one)
- Line 2431 (Istio mesh)

The most impactful fix is at **line 2128** since that's where the test fails, but applying it to all locations prevents future regressions.

### Verification

After the fix, re-run the CI workflow. Expected result: 36/36 tests pass (or more if new tests were added), with the `linkedSkills` test no longer timing out.

## Additional Context

- The `inject.kagenti.io` webhook is a mutating admission webhook that intercepts pod CREATE operations in namespaces with `kagenti.io/type`-labeled workloads.
- cert-manager provides the TLS certificate for the webhook. After a pod restart, the new pod must read the certificate from a Secret mounted as a volume. This read can take a few seconds.
- The Istio mesh test (line 2469) already uses a retry pattern for AgentRuntime creation (`3*time.Minute, 5*time.Second`) — but that alone won't fix this issue because the AgentRuntime creation succeeds; it's the subsequent pod creation by the ReplicaSet that fails.
- The controller's webhook server starts asynchronously after the main reconciliation loop. On Kind clusters in CI (resource-constrained), this startup can be slower.

## Labels

- `bug`
- `e2e`
- `priority/high`

pdettori:
## Summary

- Adds `webhookServer.StartedChecker()` as a readyz check in `cmd/main.go` so the pod Ready condition gates on the webhook TLS server actually listening on `:9443`
- Adds `ProbeWebhookReady()` utility in `test/utils/utils.go` that uses `kubectl wait --for=condition=Ready` on controller pods
- Adds the readiness probe as a second phase after the existing endpoint IP check at all 7 locations in `test/e2e/e2e_test.go`
- Fixes the race condition where the endpoint IP populates before the webhook TLS server is ready, causing pod creation failures

## Root Cause

After a controller restart, the Kubernetes `Endpoints` object gets the pod IP immediately, but the webhook's TLS server needs additional time to initialize (read certs, bind port 9443). The existing readyz check was `healthz.Ping` (a no-op), so the pod reported Ready before the webhook was accepting connections. The E2E tests only verified the endpoint IP was present, not that TLS was ready.

## Fix

1. **Runtime fix**: Add `mgr.AddReadyzCheck("webhook", webhookServer.StartedChecker())` — the controller-runtime webhook server exposes `StartedChecker()` which returns healthy only after the TLS listener is bound.
2. **Test fix**: `ProbeWebhookReady()` uses `kubectl wait --for=condition=Ready` which now reflects true webhook readiness, and handles rolling restarts (multiple pods) correctly.

## Test Plan

- [ ] CI E2E tests pass — specifically the "Skill Discovery > Feature gate enabled > should populate linkedSkills from annotation" test that was consistently failing
- [ ] No regression in other test blocks that also use the readiness probe

Fixes #420

Assisted-By: Claude Code

### DIFF
--- a/kagenti-operator/test/utils/utils.go
+++ b/kagenti-operator/test/utils/utils.go
@@ -752,21 +752,21 @@ func UncommentCode(filename, target, prefix string) error {
 	return os.WriteFile(filename, out.Bytes(), 0644)
 }
 
-// ProbeWebhookTLS runs a temporary curl pod that attempts a TLS connection to the
-// webhook service. Returns nil if the webhook is accepting TLS connections, or an
-// error if it is not yet ready. Intended for use inside an Eventually() block.
+// ProbeWebhookTLS runs a temporary curl pod that verifies the webhook service
+// TLS handshake succeeds. Uses --connect-only to avoid waiting for an HTTP
+// response (the webhook only handles admission POST requests). Returns nil if
+// the TLS connection succeeds, or an error if not yet ready.
 func ProbeWebhookTLS(namespace string) error {
 	podName := fmt.Sprintf("webhook-probe-%d", time.Now().UnixNano()%100000)
-	svcURL := fmt.Sprintf("https://kagenti-operator-webhook-service.%s.svc:443/", namespace)
+	svcURL := fmt.Sprintf("https://kagenti-operator-webhook-service.%s.svc:443", namespace)
 
-	// Security context required for clusters with "restricted" PodSecurity enforcement.
 	overrides := `{
 		"spec": {
 			"securityContext": {"runAsNonRoot": true, "seccompProfile": {"type": "RuntimeDefault"}},
 			"containers": [{
 				"name": "` + podName + `",
 				"image": "curlimages/curl:latest",
-				"command": ["curl", "-sk", "-o", "/dev/null", "-w", "%{http_code}", "--max-time", "5", "` + svcURL + `"],
+				"command": ["curl", "-sk", "--connect-only", "--connect-timeout", "5", "` + svcURL + `"],
 				"securityContext": {
 					"allowPrivilegeEscalation": false,
 					"capabilities": {"drop": ["ALL"]},
@@ -782,12 +782,9 @@ func ProbeWebhookTLS(namespace string) error {
 		"--image=curlimages/curl:latest",
 		"--overrides", overrides,
 		"-n", namespace)
-	output, err := Run(cmd)
+	_, err := Run(cmd)
 	if err != nil {
 		return fmt.Errorf("webhook TLS probe failed: %w", err)
-	}
-	if strings.TrimSpace(output) == "000" {
-		return fmt.Errorf("webhook TLS probe got HTTP 000 (connection refused)")
 	}
 	return nil
 }
--- a/kagenti-operator/test/utils/utils.go
+++ b/kagenti-operator/test/utils/utils.go
@@ -752,20 +752,18 @@ func UncommentCode(filename, target, prefix string) error {
 	return os.WriteFile(filename, out.Bytes(), 0644)
 }
 
-// ProbeWebhookReady checks that the controller pod is fully Ready (all
-// containers passing readiness probes). The manager's readiness probe gates
-// on full initialization including the webhook server startup.
+// ProbeWebhookReady checks that the controller pod is fully Ready.
+// The readyz endpoint gates on webhookServer.StartedChecker(), so Ready
+// means the webhook TLS server is accepting connections on :9443.
 func ProbeWebhookReady(namespace string) error {
-	cmd := exec.Command("kubectl", "get", "pods",
-		"-l", "control-plane=controller-manager",
+	cmd := exec.Command("kubectl", "wait",
+		"--for=condition=Ready",
+		"pod", "-l", "control-plane=controller-manager",
 		"-n", namespace,
-		"-o", "jsonpath={.items[0].status.conditions[?(@.type==\"Ready\")].status}")
-	output, err := Run(cmd)
+		"--timeout=5s")
+	_, err := Run(cmd)
 	if err != nil {
-		return fmt.Errorf("failed to check controller pod readiness: %w", err)
-	}
-	if strings.TrimSpace(output) != "True" {
-		return fmt.Errorf("controller pod not yet Ready (status=%q)", output)
+		return fmt.Errorf("controller pod not yet Ready: %w", err)
 	}
 	return nil
 }
--- a/kagenti-operator/test/utils/utils.go
+++ b/kagenti-operator/test/utils/utils.go
@@ -752,39 +752,20 @@ func UncommentCode(filename, target, prefix string) error {
 	return os.WriteFile(filename, out.Bytes(), 0644)
 }
 
-// ProbeWebhookTLS runs a temporary curl pod that verifies the webhook service
-// TLS handshake succeeds. Uses --connect-only to avoid waiting for an HTTP
-// response (the webhook only handles admission POST requests). Returns nil if
-// the TLS connection succeeds, or an error if not yet ready.
-func ProbeWebhookTLS(namespace string) error {
-	podName := fmt.Sprintf("webhook-probe-%d", time.Now().UnixNano()%100000)
-	svcURL := fmt.Sprintf("https://kagenti-operator-webhook-service.%s.svc:443", namespace)
-
-	overrides := `{
-		"spec": {
-			"securityContext": {"runAsNonRoot": true, "seccompProfile": {"type": "RuntimeDefault"}},
-			"containers": [{
-				"name": "` + podName + `",
-				"image": "curlimages/curl:latest",
-				"command": ["curl", "-sk", "--connect-only", "--connect-timeout", "5", "` + svcURL + `"],
-				"securityContext": {
-					"allowPrivilegeEscalation": false,
-					"capabilities": {"drop": ["ALL"]},
-					"runAsNonRoot": true,
-					"seccompProfile": {"type": "RuntimeDefault"}
-				}
-			}]
-		}
-	}`
-
-	cmd := exec.Command("kubectl", "run", podName,
-		"--rm", "-i", "--restart=Never",
-		"--image=curlimages/curl:latest",
-		"--overrides", overrides,
-		"-n", namespace)
-	_, err := Run(cmd)
+// ProbeWebhookReady checks that the controller pod is fully Ready (all
+// containers passing readiness probes). The manager's readiness probe gates
+// on full initialization including the webhook server startup.
+func ProbeWebhookReady(namespace string) error {
+	cmd := exec.Command("kubectl", "get", "pods",
+		"-l", "control-plane=controller-manager",
+		"-n", namespace,
+		"-o", "jsonpath={.items[0].status.conditions[?(@.type==\"Ready\")].status}")
+	output, err := Run(cmd)
 	if err != nil {
-		return fmt.Errorf("webhook TLS probe failed: %w", err)
+		return fmt.Errorf("failed to check controller pod readiness: %w", err)
+	}
+	if strings.TrimSpace(output) != "True" {
+		return fmt.Errorf("controller pod not yet Ready (status=%q)", output)
 	}
 	return nil
 }
--- a/kagenti-operator/test/utils/utils.go
+++ b/kagenti-operator/test/utils/utils.go
@@ -759,12 +759,29 @@ func ProbeWebhookTLS(namespace string) error {
 	podName := fmt.Sprintf("webhook-probe-%d", time.Now().UnixNano()%100000)
 	svcURL := fmt.Sprintf("https://kagenti-operator-webhook-service.%s.svc:443/", namespace)
 
+	// Security context required for clusters with "restricted" PodSecurity enforcement.
+	overrides := `{
+		"spec": {
+			"securityContext": {"runAsNonRoot": true, "seccompProfile": {"type": "RuntimeDefault"}},
+			"containers": [{
+				"name": "` + podName + `",
+				"image": "curlimages/curl:latest",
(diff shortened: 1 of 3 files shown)


pdettori:
## CI Status

This PR fixes the **webhook readiness race condition** (problem 1 from #420). After 4 iterations:

| Commit | Fix |
|--------|-----|
| `95057f1` | Initial: curl probe for TLS connectivity |
| `68f9a29` | Add PodSecurity context (restricted policy) |
| `3efc274` | Use `--connect-only` (webhook doesn't serve HTTP GET) |
| `1564816` | Replace curl pod with controller pod readiness check |

The final approach checks that the controller pod is fully **Ready** (readiness probe passes on port 8081), which gates on complete manager initialization including the webhook server.

### Remaining E2E failure

The Skill Discovery `linkedSkills` test still fails, but this is a **separate pre-existing issue** — the same failure occurs on `main` branch (runs `27249800636`, `27249348694`, `27249047484` all fail).

**Root cause**: The controller hits an optimistic concurrency conflict when updating `.status.linkedSkills` and does not retry. This is tracked in #422.

### What this PR fixes

- The webhook "connection refused" race after controller restart (all 7 locations)
- Prevents the cascading BeforeAll failures that were blocking all E2E tests

### What still needs fixing (separate PR)

- #422: Reconciler needs `retry.RetryOnConflict` for status updates

kagenti-operator/test/utils/utils.go
```diff
@@ -751,3 +751,21 @@ func UncommentCode(filename, target, prefix string) error {
 	// nolint:gosec
 	return os.WriteFile(filename, out.Bytes(), 0644)
 }
+
+// ProbeWebhookReady checks that the controller pod is fully Ready (all
+// containers passing readiness probes). The manager's readiness probe gates
```
huang195: **must-fix** — This comment ("the manager's readiness probe gates on full initialization including the webhook server startup") isn't accurate for this controller. In `cmd/main.go:687` the readyz check is `mgr.AddReadyzCheck("readyz", healthz.Ping)`, a no-op that returns OK as soon as the health endpoint is up. It does **not** gate on the webhook TLS server binding `:9443`. So the pod reports `Ready` *before* the webhook is accepting connections — the exact race this PR targets is not actually closed, which matches E2E still failing.

The clean fix is in `main.go`: 
```go
if err := mgr.AddReadyzCheck("webhook", webhookServer.StartedChecker()); err != nil { ... }
```
The webhook server created at `main.go:279` (`webhook.NewServer(...)`) exposes `StartedChecker()`. Once readyz actually gates on webhook startup, the pod `Ready` condition becomes meaningful and `ProbeWebhookReady` does what it claims. Alternatively, restore a real TLS connectivity probe (as the PR body still describes).

kagenti-operator/test/utils/utils.go
```diff
@@ -751,3 +751,21 @@ func UncommentCode(filename, target, prefix string) error {
...
+// on full initialization including the webhook server startup.
+func ProbeWebhookReady(namespace string) error {
+	cmd := exec.Command("kubectl", "get", "pods",
+		"-l", "control-plane=controller-manager",
+		"-n", namespace,
+		"-o", "jsonpath={.items[0].status.conditions[?(@.type==\"Ready\")].status}")
```
huang195: **suggestion** — `.items[0]` is fragile for the restart case (`e2e_test.go:2159`, "after restart"). During a rolling restart two pods coexist briefly and item ordering is nondeterministic, so this can read the *old* still-Ready pod and pass prematurely. Consider `kubectl wait --for=condition=Ready pod -l control-plane=controller-manager` (waits on all matching pods), or filter to the newest/running pod before reading the condition.

huang195 (changes requested):
## Review Summary

Right instinct, Paolo — gating the tests on real webhook readiness instead of just the endpoint IP is exactly the fix #420 needs. But the implementation rests on a premise that doesn't hold for this controller: `ProbeWebhookReady` checks the pod's `Ready` condition, and that condition is wired to `readyz = healthz.Ping` (`cmd/main.go:687`) — a no-op that flips `True` before the webhook TLS server is listening on `:9443`. So the probe reports ready too early and the underlying race survives, which is consistent with **E2E still failing** in CI (in the failed run the probe passed in ~20s, then the suite failed anyway). The real lever is one line in `main.go`: add `webhookServer.StartedChecker()` as a readyz check (the server at `main.go:279` exposes it), then this util does what it claims.

Two more things:
- **Stale PR description** — the body describes `ProbeWebhookTLS()` running a temporary curl pod with `--rm`, but the merged approach (final commit "replace curl probe with pod readiness check") is `ProbeWebhookReady` doing a `kubectl get pods` readiness check. The Test Plan's "curl probe pod cleans up correctly (`--rm`)" checkbox no longer applies — worth updating so reviewers aren't chasing the old design.
- **E2E (fail)** is the suite this PR exists to fix; it should go green before merge.

**Areas reviewed:** Go (e2e test + utils) · **Commits:** 4, all signed-off ✅ · **CI:** E2E failing (rest green, incl. pr-title ✅)

Tighten the readiness gate (`StartedChecker`) and refresh the PR body, and this lands. 🍝

Assisted-By: Claude Code

kagenti-operator/test/utils/utils.go
pdettori: Fixed in ce59ce4. Added `webhookServer.StartedChecker()` as a readyz check in `main.go:691`, so pod Ready now gates on the webhook TLS server actually listening on :9443. Updated the comment accordingly.

kagenti-operator/test/utils/utils.go
pdettori: Fixed in ce59ce4. Replaced the `.items[0]` jsonpath approach with `kubectl wait --for=condition=Ready pod -l control-plane=controller-manager` which waits on all matching pods and handles the rolling restart case correctly.

huang195 (approved):
## Review Summary (re-review)

Bravo, Paolo — this is the real fix now. Gating `readyz` on `webhookServer.StartedChecker()` (`cmd/main.go:691`) means the pod's `Ready` condition genuinely reflects the webhook TLS server accepting connections on `:9443`, so `ProbeWebhookReady` now does exactly what its (updated, accurate) comment claims. And switching the util to `kubectl wait --for=condition=Ready pod -l control-plane=controller-manager` resolves the `.items[0]` restart-case fragility cleanly — it waits on all matching pods, so the brief two-pod window during a rolling restart self-heals. Molto bene.

One leftover, non-blocking: the **PR description** still narrates the old `ProbeWebhookTLS()` curl-pod approach (and the Test Plan's `--rm` cleanup checkbox), which no longer matches the `kubectl wait` implementation — worth a refresh so it reads true.

**E2E is still running** (pending) — since closing this race is the whole point, that green is the real confirmation; worth a glance before merge.

**Areas reviewed:** Go (main.go + e2e test + utils) · **Commits:** 5, all signed-off ✅ · **CI:** all green except E2E (pending)

No blockers — approving. 🍝

Assisted-By: Claude Code

pdettori:
Needs to merge before #423 even if CI not passing as we need also #423 to pass CI.

PR #423, opened 2026-06-10
Fix: wrap status update with RetryOnConflict to prevent linkedSkills loss

### LINKED ISSUE #422: Fix: reconciler conflict error prevents linkedSkills status update
## Summary

The Skill Discovery E2E test intermittently fails because the controller's status update for `linkedSkills` hits a Kubernetes optimistic concurrency conflict and does not retry. This causes the `linkedSkills` field to never be populated, and the test times out after 180s.

**Affected**: `main` branch — all recent CI runs fail on this test.
**Impact**: 1 test fails, 5 skip (cascading from same Describe block).

## Root Cause

During reconciliation of an AgentRuntime with skill annotations, the controller attempts to update `.status.linkedSkills`. If another reconciliation loop modifies the same AgentRuntime object concurrently (e.g., adding labels, updating config-hash), the status update fails with:

```
Operation cannot be fulfilled on agentruntimes.agent.kagenti.dev "skill-discovery-agent":
the object has been modified; please apply your changes to the latest version and try again
```

The controller does not retry the status update after this conflict error. The reconciler returns the error, gets requeued, but the next reconciliation may take a different code path (object already has labels/hash), skipping the linkedSkills update.

## Evidence

From CI run `27253451563` (branch `fix/webhook-readiness-420`, June 10 2026):
- Test polls for `linkedSkills` from 04:57:42 to 05:01:20 (4+ minutes)
- Controller logs show conflict error at 04:58:20Z
- The `linkedSkills` field is never populated after the conflict

Same failure on `main` branch: runs `27249800636`, `27249348694`, `27249047484` all fail.

## Relationship to #420

Issue #420 identified two problems causing the Skill Discovery test failure:
1. **Webhook readiness race** — fixed by PR #421 (pod readiness check after restart)
2. **Reconciler conflict** — this issue (status update not retried on conflict)

PR #421 resolves problem 1. This issue tracks problem 2.

## Proposed Fix

In the Skill Discovery reconciliation path, when updating `.status.linkedSkills`:

1. **Re-fetch the latest object** before the status update (get the current resourceVersion)
2. **Use `client.Status().Patch()`** instead of `client.Status().Update()` to reduce conflict surface
3. Or: **Add retry logic** with `retry.RetryOnConflict()` from `k8s.io/client-go/util/retry`

### Example pattern

```go
import "k8s.io/client-go/util/retry"

err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
    // Re-fetch the latest version
    if err := r.Get(ctx, req.NamespacedName, agentRuntime); err != nil {
        return err
    }
    // Update status
    agentRuntime.Status.LinkedSkills = discoveredSkills
    return r.Status().Update(ctx, agentRuntime)
})
```

### Where to look

- The reconciler that handles AgentRuntime objects with skill annotations
- Search for where `linkedSkills` or `LinkedSkills` is set in the status
- The fix should wrap that status update with `retry.RetryOnConflict`

## Labels

- `bug`
- `e2e`
- `priority/high`

pdettori:
## Summary

- Wraps the reconciler's final status update (`step 8`) with `retry.RetryOnConflict` to handle optimistic concurrency conflicts
- Moves `linkedSkills` and `SkillsDiscovered` condition assignment inside the retry loop so they survive a re-fetch
- Adds a unit test simulating a concurrent status modification to verify `linkedSkills` persists through a conflict

## Root Cause

When another reconciliation loop (e.g., applying labels/config-hash) modifies the same AgentRuntime object concurrently, the status update fails with a conflict error. Without retry, the `linkedSkills` field is never populated, causing the Skill Discovery E2E test to time out after 180s.

## Approach

Uses the same `retry.RetryOnConflict(retry.DefaultRetry, ...)` pattern already established throughout this controller (see `applyWorkloadConfig`, `removeLabelsOnDelete`, etc.). Inside the retry closure:

1. Re-fetches the AgentRuntime to get the latest `resourceVersion`
2. Re-applies all status fields (configuredPods, phase, Ready condition, linkedSkills, SkillsDiscovered condition)
3. Calls `Status().Update()`

Closes #422

## Test plan

- [x] Unit test: new "should persist linkedSkills even after a concurrent status modification" test passes
- [x] Unit test: existing "should read linked skills into status" test still passes
- [ ] CI: E2E skill discovery tests pass (previously failing on every `main` run)

Assisted-By: Claude Code

### DIFF
--- a/kagenti-operator/internal/controller/agentruntime_controller.go
+++ b/kagenti-operator/internal/controller/agentruntime_controller.go
@@ -719,10 +719,6 @@ func (r *AgentRuntimeReconciler) handleDeletion(ctx context.Context, rt *agentv1
 			delete(workloadLabels, LabelManagedBy)
 			acc.obj.SetLabels(workloadLabels)
 
-			// Remove skills annotation from workload metadata.
-			workloadAnnotations := acc.obj.GetAnnotations()
-			delete(workloadAnnotations, AnnotationSkills)
-			acc.obj.SetAnnotations(workloadAnnotations)
 
 			// Remove kagenti.io/type from PodTemplateSpec pod labels so future pods
 			// are not presented to the webhook with the type label.
--- a/kagenti-operator/internal/controller/agentruntime_controller.go
+++ b/kagenti-operator/internal/controller/agentruntime_controller.go
@@ -259,27 +259,29 @@ func (r *AgentRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		logger.V(1).Info("Failed to count configured pods", "error", err)
 	}
 
-	// 8. Update status (retry on conflict to avoid losing linkedSkills updates)
+	// 8. Update status (retry on conflict to preserve all conditions computed above)
+	rt.Status.ConfiguredPods = configuredPods
+	r.setPhase(rt, agentv1alpha1.RuntimePhaseActive)
+	r.setCondition(rt, ConditionTypeReady, metav1.ConditionTrue, "Configured",
+		fmt.Sprintf("Workload %s configured with config-hash %s", rt.Spec.TargetRef.Name, configResult.Hash[:12]))
+	if fg.SkillDiscovery {
+		rt.Status.LinkedSkills = linkedSkills
+		if len(linkedSkills) > 0 {
+			r.setCondition(rt, ConditionTypeSkillsDiscovered, metav1.ConditionTrue, "SkillsFound",
+				fmt.Sprintf("%d linked skill(s) discovered from workload annotation", len(linkedSkills)))
+		} else {
+			meta.RemoveStatusCondition(&rt.Status.Conditions, ConditionTypeSkillsDiscovered)
+		}
+	} else {
+		rt.Status.LinkedSkills = nil
+		meta.RemoveStatusCondition(&rt.Status.Conditions, ConditionTypeSkillsDiscovered)
+	}
+	desired := rt.Status.DeepCopy()
 	if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		if err := r.Get(ctx, req.NamespacedName, rt); err != nil {
 			return err
 		}
-		rt.Status.ConfiguredPods = configuredPods
-		r.setPhase(rt, agentv1alpha1.RuntimePhaseActive)
-		r.setCondition(rt, ConditionTypeReady, metav1.ConditionTrue, "Configured",
-			fmt.Sprintf("Workload %s configured with config-hash %s", rt.Spec.TargetRef.Name, configResult.Hash[:12]))
-		if fg.SkillDiscovery {
-			rt.Status.LinkedSkills = linkedSkills
-			if len(linkedSkills) > 0 {
-				r.setCondition(rt, ConditionTypeSkillsDiscovered, metav1.ConditionTrue, "SkillsFound",
-					fmt.Sprintf("%d linked skill(s) discovered from workload annotation", len(linkedSkills)))
-			} else {
-				meta.RemoveStatusCondition(&rt.Status.Conditions, ConditionTypeSkillsDiscovered)
-			}
-		} else {
-			rt.Status.LinkedSkills = nil
-			meta.RemoveStatusCondition(&rt.Status.Conditions, ConditionTypeSkillsDiscovered)
-		}
+		rt.Status = *desired
 		return r.Status().Update(ctx, rt)
 	}); err != nil {
 		logger.Error(err, "Failed to update status")
--- a/kagenti-operator/internal/controller/agentruntime_controller.go
+++ b/kagenti-operator/internal/controller/agentruntime_controller.go
@@ -281,7 +281,7 @@ func (r *AgentRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		if err := r.Get(ctx, req.NamespacedName, rt); err != nil {
 			return err
 		}
-		rt.Status = *desired
+		rt.Status = *desired // safe: this controller is the sole status owner
 		return r.Status().Update(ctx, rt)
 	}); err != nil {
 		logger.Error(err, "Failed to update status")
--- a/kagenti-operator/internal/controller/agentruntime_controller.go
+++ b/kagenti-operator/internal/controller/agentruntime_controller.go
@@ -248,17 +248,9 @@ func (r *AgentRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	// 6.5. Discover linked skills from workload annotation (set by kagenti backend or user)
 	fg := r.getFeatureGates()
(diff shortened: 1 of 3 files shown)


kagenti-operator/internal/controller/agentruntime_controller.go
```diff
@@ -267,12 +259,29 @@ func (r *AgentRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
-	// 8. Update status
-	rt.Status.ConfiguredPods = configuredPods
-	r.setPhase(rt, agentv1alpha1.RuntimePhaseActive)
-	r.setCondition(rt, ConditionTypeReady, metav1.ConditionTrue, "Configured",
-		fmt.Sprintf("Workload %s configured with config-hash %s", rt.Spec.TargetRef.Name, configResult.Hash[:12]))
-	if err := r.Status().Update(ctx, rt); err != nil {
+	// 8. Update status (retry on conflict to avoid losing linkedSkills updates)
+	if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		if err := r.Get(ctx, req.NamespacedName, rt); err != nil {
```
huang195, line 281: **must-fix** — This `r.Get(ctx, req.NamespacedName, rt)` overwrites the *entire* in-memory `rt` (including `rt.Status.Conditions`) with server state. But `ConfigResolved` (set at step 5) and `IstioMeshEnrolled` (step 4.6) are set in-memory and **not yet persisted** when we reach step 8 — the re-fetch discards them, and the closure only re-applies `Ready`, `SkillsDiscovered`, `ConfiguredPods`, `LinkedSkills`. So those conditions are silently dropped on every reconcile.

This is exactly what breaks the two `agentruntime_config_test.go` specs (lines 348 & 386 — `ConfigResolved` comes back `nil`).

Suggest snapshotting the desired status before the loop and restoring it after the fresh `Get`:
```go
desired := rt.Status.DeepCopy()
if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
    if err := r.Get(ctx, req.NamespacedName, rt); err != nil {
        return err
    }
    rt.Status = *desired // keeps refreshed resourceVersion, preserves all computed conditions
    rt.Status.ConfiguredPods = configuredPods
    // ... existing Ready / LinkedSkills / SkillsDiscovered logic ...
    return r.Status().Update(ctx, rt)
}); err != nil { ... }
```

kagenti-operator/internal/controller/agentruntime_controller_test.go
```diff
@@ -209,6 +209,44 @@ var _ = Describe("AgentRuntime Controller", func() {
...
+			Expect(k8sClient.Status().Update(ctx, current)).To(Succeed())
+
+			// Now reconcile — the status update inside Reconcile will initially
+			// hit a conflict (stale resourceVersion), but retry should succeed
+			_, err := r.Reconcile(ctx, reconcile.Request{NamespacedName: nn})
+			Expect(err).NotTo(HaveOccurred())
+
+			updatedRT := &agentv1alpha1.AgentRuntime{}
+			Expect(k8sClient.Get(ctx, nn, updatedRT)).To(Succeed())
+			Expect(updatedRT.Status.LinkedSkills).To(ConsistOf("skill-a", "skill-b"))
```
huang195, line 248: **suggestion** — This new test only asserts `LinkedSkills` survives the conflict, which is why it stayed green while the `ConfigResolved` / `IstioMeshEnrolled` regression slipped through. Adding an assertion that `ConfigResolved` (and ideally `IstioMeshEnrolled`) is still present after the conflicting reconcile would lock down the whole status-preservation contract, not just the one field.

huang195 (changes requested):
## Review Summary

The diagnosis and the `RetryOnConflict` approach are spot-on, Paolo — bellissimo reuse of the established controller pattern, and the root-cause writeup is chef's-kiss. The catch: re-fetching `rt` inside the retry closure throws away the conditions set earlier in the same reconcile that weren't persisted yet, so `linkedSkills` is saved but `ConfigResolved` and `IstioMeshEnrolled` are lost. The unit-test suite caught it (2 failing specs in a file this PR doesn't even touch — that's the tell).

Two CI gates to clear:
- **Unit Tests (fail)** — the condition-drop regression (see inline). The real blocker.
- **verify-pr-title (fail)** — title must use a **capitalized** prefix. `fix:` → **`Fix:`** (the check is case-sensitive: `Build, Chore, CI, Docs, Feat, Fix, ...`).
- **E2E Tests (fail)** — this is the suite the PR targets; worth confirming it goes green once the condition-drop is fixed (the test-plan checkbox for it is still unchecked).

**Areas reviewed:** Go (controller + test) · **Commits:** 1, signed-off ✅ · **CI:** failing (Unit, E2E, pr-title)

Fix the re-fetch to preserve all conditions and that title prefix, and this is good to go. 🍝

Assisted-By: Claude Code

kagenti-operator/internal/controller/agentruntime_controller.go
```diff
@@ -267,12 +259,31 @@ func (r *AgentRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	} else {
+		rt.Status.LinkedSkills = nil
+		meta.RemoveStatusCondition(&rt.Status.Conditions, ConditionTypeSkillsDiscovered)
+	}
+	desired := rt.Status.DeepCopy()
+	if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		if err := r.Get(ctx, req.NamespacedName, rt); err != nil {
+			return err
+		}
+		rt.Status = *desired
```
huang195: **nit** — `rt.Status = *desired` is last-writer-wins on the *entire* status: any field a concurrent actor wrote that this reconcile didn't recompute gets overwritten. That's correct here since this controller is the sole owner of `AgentRuntime` status, but a one-line comment noting that intent would save a future reader from mistaking it for an oversight.

huang195 (approved):
## Review Summary (re-review)

Perfetto, Paolo — this is exactly the fix. Snapshotting `rt.Status.DeepCopy()` before the retry loop and restoring it after the fresh `Get` preserves `ConfigResolved`, `IstioMeshEnrolled`, and the rest while still getting the refreshed `resourceVersion` for a conflict-free update. And bellissimo touch adding the `ConfigResolved`-survives-the-retry assertion to the new test — that locks down the whole status-preservation contract, not just `linkedSkills`. Unit Tests, pr-title, and lint are all green now.

Only open item: **E2E is still running** (pending) — since that's the suite this PR ultimately targets (skill discovery timeout), worth a glance once it lands before merge.

**Areas reviewed:** Go (controller + test) · **Commits:** 2, all signed-off ✅ · **CI:** all green except E2E (pending)

No blockers — approving. 🍝

Assisted-By: Claude Code

kevincogan (approved):
LGTM and E2E is green. Nice work!

PR #425, opened 2026-06-10
Fix(mlflow): add --mlflow-ca-file flag for custom TLS CA verification

### LINKED ISSUE #1896: E2E: weather-tool fails to verify MLflow TLS cert (x509: certificate signed by unknown authority)
## Summary

The weather-tool agent cannot connect to the MLflow tracing endpoint because it doesn't trust the MLflow server's TLS certificate. This causes `test_agent_simple_query` and `test_agent_multiturn_conversation` to fail with MCP connectivity errors.

## Error

```
Cannot connect to MCP weather service at http://weather-tool-mcp.team1.svc.cluster.local:8000/mcp
```

Root cause in weather-tool logs:
```
tls: failed to verify certificate: x509: certificate signed by unknown authority
```

The weather-tool pod is trying to send traces to MLflow over HTTPS but doesn't have the cluster CA bundle mounted.

## Scope

- Affects **all** HyperShift E2E runs (main branch fails identically — run 27297132003)
- Pre-existing issue, not introduced by any recent PR
- Separate from #1890 (SharedTrustReconciler / istiod BadSignature), which is fixed by #1892

## Potential Fixes

1. Mount the cluster CA bundle (`/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` or OpenShift's trusted-ca-bundle) into the weather-tool pod
2. Configure the OTEL exporter in weather-tool to skip TLS verification for in-cluster endpoints
3. Use HTTP instead of HTTPS for in-cluster MLflow trace export (service-to-service within the mesh)

## Reproduction

Any `/run-e2e` on any PR — the two agent conversation tests fail consistently.

Assisted-By: Claude Code

pdettori:
## Summary

- Add `--mlflow-ca-file` flag to the operator binary that configures the MLflow REST client with a custom CA bundle
- The CA PEM is appended to the system cert pool, so both publicly-trusted and cluster-specific CAs work
- Fixes TLS verification failure on HyperShift clusters where the RHOAI MLflow gateway uses the cluster ingress CA

Closes kagenti/kagenti#1896

## Context

On HyperShift CI clusters, the operator's `MLflowReconciler` fails to create experiments:
```
Post "https://rh-ai.apps.*.*/mlflow/api/2.0/mlflow/experiments/create":
tls: failed to verify certificate: x509: certificate signed by unknown authority
```

The MLflow gateway URL (from `mlflows.mlflow.opendatahub.io` CR `.status.url`) uses an OpenShift Route whose TLS certificate is signed by the cluster's ingress CA — not a publicly-trusted CA.

## Changes

| File | Change |
|------|--------|
| `internal/mlflow/client.go` | Add `CAFile` field and `buildTransport()` to load custom CA |
| `internal/controller/mlflow_controller.go` | Add `MLflowCAFile` field, pass to client |
| `cmd/main.go` | Add `--mlflow-ca-file` flag, wire to reconciler |
| `internal/mlflow/client_test.go` | Unit tests for `buildTransport` |

## Test plan

- [x] Unit tests pass (`go test ./internal/mlflow/`)
- [x] Build succeeds (`go build ./...`)
- [ ] Integration: deploy with `--mlflow-ca-file=/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` on HyperShift
- [ ] Companion PR in kagenti/kagenti mounts the trusted-ca-bundle ConfigMap

Assisted-By: Claude Code

### DIFF
--- a/kagenti-operator/internal/mlflow/client.go
+++ b/kagenti-operator/internal/mlflow/client.go
@@ -158,36 +158,50 @@ func (c *Client) retryBaseDelay() time.Duration {
 	return DefaultRetryBaseDelay
 }
 
-func (c *Client) httpClient() *http.Client {
+func (c *Client) httpClient() (*http.Client, error) {
+	var initErr error
 	c.httpOnce.Do(func() {
 		if c.HTTPClient == nil {
+			transport, err := c.buildTransport()
+			if err != nil {
+				initErr = err
+				return
+			}
 			c.HTTPClient = &http.Client{
 				Timeout:   c.timeout(),
-				Transport: c.buildTransport(),
+				Transport: transport,
 			}
 		}
 	})
-	return c.HTTPClient
+	if initErr != nil {
+		return nil, initErr
+	}
+	if c.HTTPClient == nil {
+		return nil, fmt.Errorf("http client initialization failed")
+	}
+	return c.HTTPClient, nil
 }
 
-func (c *Client) buildTransport() http.RoundTripper {
+func (c *Client) buildTransport() (http.RoundTripper, error) {
 	if c.CAFile == "" {
-		return nil
+		return nil, nil
 	}
 	caPEM, err := os.ReadFile(c.CAFile)
 	if err != nil {
-		return nil
+		return nil, fmt.Errorf("reading CA file %q: %w", c.CAFile, err)
 	}
 	pool, err := x509.SystemCertPool()
 	if err != nil {
 		pool = x509.NewCertPool()
 	}
-	pool.AppendCertsFromPEM(caPEM)
+	if !pool.AppendCertsFromPEM(caPEM) {
+		return nil, fmt.Errorf("CA file %q contains no valid PEM certificates", c.CAFile)
+	}
 	return &http.Transport{
 		TLSClientConfig: &tls.Config{
 			RootCAs: pool,
 		},
-	}
+	}, nil
 }
 
 func (c *Client) tokenPath() string {
@@ -398,7 +412,12 @@ func (c *Client) doRequest(ctx context.Context, method, path, workspace string, 
 		req.Header.Set("Content-Type", "application/json")
 	}
 
-	resp, err := c.httpClient().Do(req)
+	httpClient, err := c.httpClient()
+	if err != nil {
+		return nil, fmt.Errorf("initializing HTTP client: %w", err)
+	}
+
+	resp, err := httpClient.Do(req)
 	if err != nil {
 		return nil, fmt.Errorf("executing request: %w", err)
 	}
--- a/kagenti-operator/internal/mlflow/client.go
+++ b/kagenti-operator/internal/mlflow/client.go
@@ -22,6 +22,8 @@ package mlflow
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


kagenti-operator/internal/mlflow/client.go
```diff
@@ -154,13 +162,34 @@ func (c *Client) httpClient() *http.Client {
...
 		}
 	})
 	return c.HTTPClient
 }
 
+func (c *Client) buildTransport() http.RoundTripper {
+	if c.CAFile == "" {
+		return nil
+	}
+	caPEM, err := os.ReadFile(c.CAFile)
```
mrsabath, line 189: **suggestion** — When `CAFile` is set but `os.ReadFile` fails (missing/unreadable file), `buildTransport` returns `nil`, so the client silently falls back to the system pool. On a cluster that needs the custom CA, that resurfaces as exactly the `x509: certificate signed by unknown authority` error this PR is fixing — except now it's masking a config error rather than a missing-CA error. Since the package has no logger and this returns an `http.RoundTripper`, the cleanest fix is probably to have `buildTransport` return `(http.RoundTripper, error)` and surface it at client-construction time (or log it in the controller, which does have a logger). Not a blocker, but a configured-but-broken CA path failing loudly would save a real debugging session.

kagenti-operator/internal/mlflow/client.go
```diff
@@ -154,13 +162,34 @@ func (c *Client) httpClient() *http.Client {
...
+	if err != nil {
+		return nil
+	}
+	pool, err := x509.SystemCertPool()
+	if err != nil {
+		pool = x509.NewCertPool()
+	}
+	pool.AppendCertsFromPEM(caPEM)
```
mrsabath: **suggestion** — `pool.AppendCertsFromPEM(caPEM)` returns a bool that's ignored here. A malformed or empty PEM appends zero certs and you'd get a transport with no added trust root — again surfacing as a confusing x509 failure rather than a clear "CA file contained no valid certificates." Worth checking the return and treating `false` the same as a load failure (see the note above).

kagenti-operator/internal/mlflow/client_test.go
```diff
@@ -431,3 +431,50 @@ func TestRetry_NoRetryOn4xx(t *testing.T) {
...
+	c := &Client{BaseURL: "https://example.com", CAFile: caPath}
+	transport := c.buildTransport()
+	if transport == nil {
+		t.Fatal("expected non-nil transport with valid CA file")
+	}
+	httpTransport, ok := transport.(*http.Transport)
+	if !ok {
+		t.Fatal("expected *http.Transport")
+	}
+	if httpTransport.TLSClientConfig == nil || httpTransport.TLSClientConfig.RootCAs == nil {
```
mrsabath, line 494: **nit** — `TestBuildTransport_ValidCAFile` only asserts `RootCAs != nil`, but `x509.SystemCertPool()` already returns a non-nil pool — so this passes even if `AppendCertsFromPEM` rejected the dummy cert. To actually exercise the append path, compare subject counts before/after (`len(pool.Subjects())`) or generate a real self-signed cert with `crypto/x509` + `crypto/ecdsa` in the test rather than a hardcoded PEM blob.

mrsabath (approved):
## Summary

Clean, focused fix. Adds `--mlflow-ca-file` so the MLflow REST client trusts a custom CA appended to the system pool — correctly keeps TLS verification on (no `InsecureSkipVerify`), preserves default-transport behavior when the flag is unset, and is well covered by the happy-path unit tests. All 17 CI checks pass and the commit is signed off.

Approving. The three inline notes are all non-blocking and circle one theme: when a CA file *is* configured but can't be loaded, the failure is swallowed and resurfaces as the very same confusing x509 error this PR exists to fix. Worth a follow-up but not a merge blocker.

**Areas reviewed**: Go (client, controller, cmd, tests), security (TLS), tests
**Commits**: 1, signed-off: yes
**CI status**: passing (17/17)

_Bel lavoro, Paolo — il giro lungo del certificato è finalmente chiuso._

PR #428, opened 2026-06-12
Chore: Retire CLUSTER_CIDRS injection (DNS exemption is resolv.conf-driven)

huang195:
## Summary

Retire the `CLUSTER_CIDRS` knob from the operator now that proxy-init derives the DNS exemption from the pod's actual resolvers.

[kagenti-extensions#499](https://github.com/kagenti/kagenti-extensions/pull/499) changes the `enforce-redirect` egress guard to leave cluster DNS direct by exempting the `/etc/resolv.conf` `nameserver` IPs (`UDP/53` + `TCP/53`), instead of a guessed in-cluster CIDR. The old `CLUSTER_CIDRS` default was Kind-shaped `10.0.0.0/8` and **silently dropped DNS on OpenShift/HyperShift**, where the resolver lives in the service network (`172.30`/`172.31`, outside `10/8`) — the root cause of the kagenti#1911 e2e failure. The resolver is also unpredictable in general (any service ClusterIP, or a NodeLocal DNSCache at link-local `169.254.x`), so no CIDR guess covers it.

The init script no longer reads `CLUSTER_CIDRS`, making this operator wiring dead code.

## Changes

- **`container_builder.go`** — stop injecting the `CLUSTER_CIDRS` env on the `enforce-redirect` proxy-init container.
- **`config/types.go`** — drop `ProxyConfig.ClusterCIDRs`, its deep-copy, and the `Validate()` non-empty + CIDR-syntax checks (plus the now-unused `net` import).
- **`config/defaults.go`** — drop the `10.0.0.0/8` default.
- **`charts/kagenti-operator/values.yaml`** — drop `proxy.clusterCIDRs`; the comment now notes DNS is resolv.conf-driven with no per-cluster config.
- **tests** — remove `TestValidate_ClusterCIDRs`; the enforce-redirect container test now asserts `CLUSTER_CIDRS` is **not** set.

## Sequencing / compatibility

Safe to land independently: the updated init script ignores `CLUSTER_CIDRS` even if an older operator still injects it, and this operator drops the injection — neither requires the other. The end state needs both ext#499 (new proxy-init image) and this PR.

`go build` / `go vet` / `go test ./internal/webhook/...` all green.

### DIFF
--- a/charts/kagenti-operator/values.yaml
+++ b/charts/kagenti-operator/values.yaml
@@ -225,13 +225,9 @@ defaults:
     # always-on enforce-redirect egress guard (proxy-sidecar / lite). MUST match
     # the authbridge listener.transparent_proxy_addr (default :8082).
     transparentPort: 8082
-    # In-cluster CIDRs (pods/services/DNS) left direct by the enforce-redirect
-    # guard; external TCP is REDIRECTed to the transparent listener and external
-    # non-TCP is dropped. The default is Kind-shaped (pods 10.244/16 + services
-    # 10.96/16). OpenShift/EKS MUST override — e.g. OCP services 172.30.0.0/16
-    # (outside 10/8) + pods 10.128.0.0/14 — or in-cluster service traffic breaks.
-    clusterCIDRs:
-      - "10.0.0.0/8"
+    # Cluster DNS is kept direct by proxy-init itself (it reads the pod's
+    # /etc/resolv.conf nameservers), so there is no in-cluster CIDR knob to set —
+    # works on Kind / OpenShift / EKS / NodeLocal-DNSCache with no per-cluster config.
 
   # Resource defaults (conservative for dev)
   # Note: requests must be <= limits
--- a/kagenti-operator/internal/webhook/config/defaults.go
+++ b/kagenti-operator/internal/webhook/config/defaults.go
@@ -35,9 +35,6 @@ func CompiledDefaults() *PlatformConfig {
 			// Transparent listener port — must match the authbridge proxy-sidecar
 			// preset (listener.transparent_proxy_addr default :8082).
 			TransparentPort: 8082,
-			// Kind-shaped default (pods 10.244/16 + services 10.96/16). OCP/EKS
-			// MUST override (see ProxyConfig.ClusterCIDRs doc).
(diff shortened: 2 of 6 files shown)


mrsabath (approved):
## Review Summary

Clean operator-side follow-up to [ext#499](https://github.com/kagenti/kagenti-extensions/pull/499). Retires the now-dead `CLUSTER_CIDRS` wiring: stops injecting the env var, drops `ProxyConfig.ClusterCIDRs` (+ deep-copy, `Validate()` CIDR checks, the unused `net` import), removes the `10.0.0.0/8` default and the `proxy.clusterCIDRs` chart value, and flips the container test to assert the var is **absent**. DNS exemption is now resolv.conf-driven in proxy-init, so this wiring is genuinely dead.

### Verification performed
- ✅ **No leftover references**: grepped the entire PR-head tree — the only remaining `CLUSTER_CIDRS` hits are the test comment + the negative assertion. No live consumer remains, and no chart/CRD/sample YAML still sets `clusterCIDRs`.
- ✅ **Import removals correct**: `net` has zero remaining usages in `types.go`; `strings` has zero remaining usages in `container_builder_test.go` (and is correctly retained in `container_builder.go`, still used). Build/vet pass — corroborated by green Build / Lint / Unit / Integration CI.
- ✅ **values.yaml** stays valid (key + list item fully removed, no dangling key; Helm + YAML lint green).
- ✅ **Sequencing claim holds**: consistent with ext#499 — the init script ignores `CLUSTER_CIDRS` even if injected, so either side lands independently; the end state needs both.
- ✅ **Test quality**: the negative assertion (`if _, ok := got["CLUSTER_CIDRS"]; ok`) is the right way to lock in the removal, and the obsolete `TestValidate_ClusterCIDRs` is correctly dropped.

**Areas reviewed**: Go, Helm, Tests. **Commit**: 1, signed-off (DCO green), conventional `Chore:` format. **CI**: green except E2E pending (non-blocking; unit + integration cover this config-removal change). **Verdict**: APPROVE — no blocking issues, nothing to nit.

善始善终 (shàn shǐ shàn zhōng — "finish well what you started well"): shipping the operator cleanup right behind the proxy-init fix instead of leaving dead wiring around. 干净利落。

PR #429, opened 2026-06-15
Fix: address security audit findings across operator codebase

varshaprasad96:
## Summary

Addresses HIGH+ severity findings from the RHOAI security audit report (agents-operator, 2026-06-11). Fixes are scoped to code that exists in kagenti-operator — authbridge findings are upstream-only and not in scope.

### Security fixes (5 genuine issues)
- **Pin GitHub Action to SHA** — `.github/pr-verifier.yaml` uses `pull_request_target` with a mutable tag; pinned to commit SHA to prevent supply chain attacks
- **Separate secrets/configmaps RBAC** — split into distinct rules for auditability; aligned Helm chart `delete` verb with kustomize
- **Add securityContext to sample manifest** — `runAsNonRoot`, `seccompProfile: RuntimeDefault`, drop all capabilities
- **Fix StatefulSet workload name derivation** — `deriveWorkloadInfo` now uses controller ownerReferences to correctly identify StatefulSet pods (previously returned `"myapp-0"` instead of `"myapp"`, so AgentRuntime overrides never matched)
- **Extend MLflow namespace discovery** — `agentNamespaces()` now lists both Deployments and StatefulSets with agent labels

### Determinism fixes (3 defense-in-depth)
- **readClusterFeatureGates** — try well-known `feature-gates.yaml` key first, then sorted fallback (was iterating Go map nondeterministically)
- **extractTrustDomainFromBundle** — sort trust domains and pick lexicographically smallest; also fixed the caller `discoverFromSpireBundle` which had the same nondeterministic iteration
- **ReadAgentRuntimeOverrides** — match on `targetRef.Kind` + `Name` (was name-only); logs a warning when kind is unknown and multiple CRs match

### Performance & quality
- **ConfigLoader + FeatureGateLoader** — replaced `RWMutex` + `DeepCopy`-on-`Get()` with `atomic.Pointer` for lock-free reads on the webhook admission hot path
- **Shared ownership helper** — extracted pod→workload resolution into `internal/workload` package, removing duplicated logic between webhook and controller
- **PlatformConfig immutability** — documented the read-only contract on the struct type

### Review-driven improvements
After the initial fixes, a high-effort 7-angle code review found 10 issues in the changes themselves:
- Fixed dead `"globalEnabled"` preferred-key logic in `readClusterFeatureGates`
- Added `ref.Controller == true` guard on StatefulSet OwnerReference check
- Added test coverage for kind-mismatch, empty-kind fallback, and `workload.ResolveOwner`
- Fixed nondeterministic map iteration in `discoverFromSpireBundle` caller
- Aligned `FeatureGateLoader` with the same `atomic.Pointer` pattern as `ConfigLoader`

## Test plan
- [x] All unit tests pass (`go test ./...` — only e2e skipped, requires Docker)
- [x] New tests for `workload.ResolveOwner`, `IsPodOwnedBy` (7 cases)
- [x] New tests for `ReadAgentRuntimeOverrides` kind-mismatch and empty-kind paths (2 cases)
- [ ] Verify sidecar injection works for Deployment-backed agents on a dev cluster
- [ ] Verify sidecar injection works for StatefulSet-backed agents on a dev cluster
- [ ] Verify MLflow OTEL RoleBindings are created for StatefulSet agent namespaces

Signed-off-by: Varsha Prasad Narsing <varshaprasad96@gmail.com>


varshaprasad96:
@akram @Ygnas - opened this PR based on security scan. Could you review and lmk your thoughts.

kagenti-operator/internal/webhook/config/loader.go
```diff
@@ -90,13 +94,11 @@ func (l *ConfigLoader) Load() error {
 	return nil
 }
 
-// Get returns current config (thread-safe)
+// Get returns the current config snapshot (lock-free, no copy).
+// The returned pointer must be treated as read-only; callers that need
+// to mutate config should DeepCopy explicitly.
 func (l *ConfigLoader) Get() *PlatformConfig {
```
pdettori, line 106: **suggestion** — `Get()` now returns the shared singleton pointer instead of a DeepCopy. `PlatformConfig` carries reference-typed fields (notably `Resources` → `corev1.ResourceRequirements` with `Limits`/`Requests` maps), and `ContainerBuilder` assigns e.g. `b.cfg.Resources.EnvoyProxy` directly into pod container specs. The struct copy shares the underlying maps with the singleton, so the read-only contract now depends on nothing downstream mutating those aliased maps/slices.

I verified no current caller does — so this is safe today — but the old DeepCopy-on-`Get()` masked this entirely. Worth a one-line note here (the `types.go` doc-comment says treat-as-read-only, which is good) so a future caller assigning a resource map into a spec and then tweaking it doesn't silently corrupt the shared config. The `FeatureGates` comment correctly notes it's value-only; `PlatformConfig` is the one with the reference-type sharp edge.

kagenti-operator/internal/discovery/discovery.go
```diff
@@ -136,15 +145,22 @@ func discoverFromSpireBundle(ctx context.Context, c client.Reader, ns string) (s
...
-			return domain, nil
+			domains = append(domains, domain)
 		}
 	}
-	return "", fmt.Errorf("empty bundle")
+	if len(domains) == 0 {
+		return "", fmt.Errorf("empty bundle")
+	}
+	sort.Strings(domains)
+	return domains[0], nil
```
pdettori, line 169: **suggestion** — returning the lexicographically-smallest trust domain makes this deterministic (✓ the goal), but if a SPIRE bundle ConfigMap ever federates multiple trust domains, "smallest name" could pick a federated peer's domain over the local one. Consider documenting the single-domain assumption, or selecting the domain more intentionally (e.g. the one matching the namespace/expected local TD) rather than by sort order.

kagenti-operator/internal/workload/owner.go
```diff
@@ -0,0 +1,91 @@
...
+	if pod.Name != "" {
+		return OwnerInfo{Name: pod.Name, Kind: ""}
+	}
+	return OwnerInfo{}
+}
+
+// IsPodOwnedBy returns true if the pod is owned (directly or via a
+// ReplicaSet) by the named workload. Supports Deployment (via
+// ReplicaSet), StatefulSet, and Sandbox ownership chains.
+func IsPodOwnedBy(pod *corev1.Pod, workloadName string) bool {
```
pdettori, line 68: **nit** — `IsPodOwnedBy` doesn't apply the `ref.Controller == true` guard that `ResolveOwner` enforces a few lines up. This preserves the pre-refactor behavior so it's not a bug, but having two ownership helpers in the same file with different controller-ref semantics is a subtle footgun. Either a comment noting the intentional difference, or aligning both on controller-only refs, would help the next reader.

pdettori (approved):
## Review Summary

Solid, well-scoped security-audit PR. I verified the higher-risk changes rather than taking them on faith:

- **Action SHA pin** — `012269a88fa4c034a0acf1ba84c26b195c0dbab4` confirmed == `refs/tags/v0.4.3` of `kubebuilder-release-tools`. Correct pin.
- **`atomic.Pointer` refactor** — traced every consumer of `ConfigLoader.Get()` / `FeatureGateLoader.Get()` (`ResolveConfig`, `ContainerBuilder`, precedence evaluator). All treat the returned config/gates as read-only — no field writes in the webhook path — so the lock-free swap is safe. (See inline note on reference-typed fields for a forward-looking caveat.)
- **StatefulSet ownership fix** (`ResolveOwner`) — the controller-ownerRef guard is correct and genuinely fixes the `myapp-0` → `myapp` derivation bug.
- **`ReadAgentRuntimeOverrides`** kind+name matching — loop is correct, backward-compatible when kind is empty, and well tested.
- Determinism fixes (feature-gates key preference, SPIRE bundle ordering) read correctly.

**Author**: varshaprasad96 (MEMBER — maintainer)
**Areas reviewed**: Go (controllers, webhook, discovery, workload), RBAC (Helm + kustomize), CI action pinning, sample manifest securityContext
**Agent/IDE config (.claude/.vscode)**: none
**CI status**: all green except **E2E Tests**, which failed with `panic: test timed out after 30m0s` while *pulling/loading AuthBridge sidecar images into Kind* — an infra/image-load timeout flake, not a code regression (no image-ref or e2e-setup changes in this PR; `main` is green). Recommend a re-run.

**RBAC note**: the Helm chart now grants `delete` on `secrets`, widening it to match the kustomize-generated role (source of truth from kubebuilder markers). Justified by operator-managed Keycloak client-credential cleanup; flagging only because it's a privilege widening in a security PR.

Approving — the inline items are suggestions/nits, not blockers.

PR #432, opened 2026-06-16
Feat: Expose configurable iptables backend (IPTABLES_CMD) for proxy-init

huang195:
## What

Exposes an optional, configurable iptables backend for the injected `proxy-init`
container. New `ProxyConfig.IptablesCmd` is injected as the `IPTABLES_CMD` env
var (both `redirect` and `enforce-redirect` modes), **omitted when empty**.

## Why

Companion to **kagenti/kagenti-extensions#518**, which switches `proxy-init`'s
`detect_iptables_cmd()` to a `/proc/modules` functional check (`iptable_nat`
loaded → legacy, as on Kind/kubeadm; absent → nft, as on OpenShift/ROSA) and
adds loud-fail on misprogram. That script change is self-sufficient — it
auto-detects correctly without operator input. This PR adds the **deterministic
per-platform override**: an operator/admin can pin the backend (e.g.
`iptables` on nft-only platforms) where auto-detection is wrong or undesired,
without rebuilding the image.

Default is empty, so behavior is unchanged unless explicitly set.

## Changes

- `config`: `ProxyConfig.IptablesCmd` (json/yaml tags), empty compiled default, logged in `logConfig`.
- `injector`: append `IPTABLES_CMD` env only when non-empty, after the mode switch (covers both modes); default injection unchanged.
- `chart`: `defaults.proxy.iptablesCmd: ""` in `values.yaml`, documented.
- `tests`: `IPTABLES_CMD` absent by default; present with the configured value in both modes.

## Testing

- `go build ./...`, `go vet ./internal/webhook/...`, `go test ./internal/webhook/{config,injector}/...` — all pass (incl. the new `TestBuildProxyInitContainer_IptablesCmd`).
- `gofmt` clean; `helm lint charts/kagenti-operator` passes.
- golangci-lint: no new findings in changed files. (Pre-existing findings in untouched `pod_mutator.go` / `volume_builder_test.go` are out of scope.)

## Notes

- This is **not** a ROSA enablement on its own — see extensions#518. On ROSA, selecting nft still hits the `container_t` → nf_tables SELinux denial; the SELinux/SCC track is separate (kagenti-extensions#502).
- A follow-up could auto-default this to `iptables` when the operator detects OpenShift (detection already exists in `internal/bootstrap/otel.go`), but it's intentionally left as an explicit knob here since the script auto-detects.

Refs kagenti/kagenti-extensions#518, kagenti/kagenti-extensions#502

### DIFF
--- a/charts/kagenti-operator/values.yaml
+++ b/charts/kagenti-operator/values.yaml
@@ -228,6 +228,11 @@ defaults:
     # Cluster DNS is kept direct by proxy-init itself (it reads the pod's
     # /etc/resolv.conf nameservers), so there is no in-cluster CIDR knob to set —
     # works on Kind / OpenShift / EKS / NodeLocal-DNSCache with no per-cluster config.
+    # iptables backend override, injected as IPTABLES_CMD. Empty (default) lets
+    # proxy-init auto-detect from /proc/modules (iptable_nat loaded => legacy, as
+    # on Kind/kubeadm; absent => nft, as on OpenShift/ROSA). Set to "iptables"
+    # (nft) or "iptables-legacy" to force a backend where detection is undesired.
+    iptablesCmd: ""
 
   # Resource defaults (conservative for dev)
   # Note: requests must be <= limits
--- a/kagenti-operator/internal/webhook/config/defaults.go
+++ b/kagenti-operator/internal/webhook/config/defaults.go
@@ -35,6 +35,9 @@ func CompiledDefaults() *PlatformConfig {
 			// Transparent listener port — must match the authbridge proxy-sidecar
 			// preset (listener.transparent_proxy_addr default :8082).
 			TransparentPort: 8082,
+			// Empty by default: proxy-init auto-detects the iptables backend from
+			// /proc/modules. Set (e.g. "iptables") to force a backend per-platform.
+			IptablesCmd: "",
 		},
 		Resources: ResourcesConfig{
 			EnvoyProxy: corev1.ResourceRequirements{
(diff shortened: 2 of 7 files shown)


esnible (approved):
Clean, well-scoped PR. Adds an optional `ProxyConfig.IptablesCmd` knob injecting `IPTABLES_CMD` into the proxy-init container (both modes, omitted when empty) as the deterministic operator-side companion to the `/proc/modules` auto-detection in kagenti-extensions#518.

Verified:
- **End-to-end wiring**: `values.yaml` `defaults.proxy.iptablesCmd` → `configmap-platform-defaults.yaml` (`.Values.defaults | toYaml`) → `yaml.Unmarshal` (loader honors the new json/yaml tags) → `ProxyConfig.IptablesCmd` → env injection.
- **Validation whitelist matches real binaries**: `iptables`, `iptables-nft`, `iptables-legacy` all exist in the proxy-init image (confirmed `iptables-nft` ships in the Alpine 3.23 `iptables` package at `/usr/sbin/iptables-nft`; `iptables-legacy` via its own package per ext-518 Dockerfile). The Validate() check genuinely prevents the per-injected-pod init crash it documents.
- **Script contract**: ext-518's `detect_iptables_cmd()` short-circuits on a non-empty `IPTABLES_CMD`, else falls back to `/proc/modules`. Matches.
- **Dual-mode parity**: `IPTABLES_CMD` appended after the mode switch, same config source for both `redirect` and `enforce-redirect`; the new test asserts this across both modes.
- **Tests** are assertive: default-absent, configured-present (both modes), and validation accept/reject incl. typo rejection.

All 16 CI checks pass; both commits signed off and use `Assisted-By`. No blocking, suggestion, or nit issues.

Assisted-By: Claude Code

PR #434, opened 2026-06-16
Fix: preserve explicitly-set authbridge-config keys per namespace (#433)

### LINKED ISSUE #433: AuthbridgeConfigReconciler overwrites per-namespace KEYCLOAK_REALM (regression in cdbed3b) — breaks non-default realms / token-exchange e2e
## Summary

`AuthbridgeConfigReconciler` unconditionally overwrites the `KEYCLOAK_REALM` key (and other Keycloak keys) of every watched namespace's `authbridge-config` ConfigMap with the **platform-global** realm on each reconcile. This silently discards any explicitly-set per-namespace realm, effectively removing the ability to run workloads against a non-default realm. It is a regression introduced by `cdbed3b` ("feat: operator-driven identity deployment").

## Impact

Any namespace that needs a realm other than the platform default can no longer get one: even if an operator/admin/CI sets `authbridge-config.KEYCLOAK_REALM=<other>`, the reconciler resets it to `Platform.KeycloakRealm` (default `"kagenti"`) within one reconcile cycle. The client-registration controller then registers the workload client in the wrong realm.

Concretely, this breaks the kagenti repo's token-exchange E2E suite (`E2E K8s 1.35.0 (Kind)` push workflow), which has been red continuously as a result. The first test `TestClientCredentials` fails with:

```
client_credentials failed: {"error":"invalid_client","error_description":"Invalid client or Invalid client credentials"}
assert 401 == 200
```

## Root cause

On `main`, `internal/controller/authbridgeconfig_controller.go`:

```go
// Update existing ConfigMap
...
for k, v := range desired {          // ~line 114 — overwrites existing keys
    cm.Data[k] = v
}
```

where `desired` is built from platform config:

```go
realm := r.Platform.KeycloakRealm    // ~line 132
if realm == "" {
    realm = defaultKeycloakRealm     // "kagenti"  (line 28)
}
...
"KEYCLOAK_REALM": realm,             // ~line 159
```

The client-registration controller then reads that overwritten value:

```go
KeycloakRealm: cm.Data["KEYCLOAK_REALM"],   // clientregistration_controller.go:357
```

So a workload in namespace `tx-e2e` (where the harness created realm `tx-e2e` and set `KEYCLOAK_REALM=tx-e2e`) gets its client registered in realm `kagenti` instead. A `client_credentials` grant against realm `tx-e2e` then returns `invalid_client`.

## Reproduction

1. Build/run the operator from `main`.
2. Create a Keycloak realm `tx-e2e`.
3. In a watched namespace, `kubectl apply` an `authbridge-config` ConfigMap with `KEYCLOAK_REALM: tx-e2e`.
4. Observe (within ~one reconcile) that `kubectl get cm authbridge-config -n <ns> -o jsonpath='{.data.KEYCLOAK_REALM}'` reverts to `kagenti`.
5. The registered client (`kagenti-keycloak-client-credentials-*`, `client-id.txt`) lands in realm `kagenti`, not `tx-e2e`.

## Version note

This regression is **not present in `v0.2.0-rc.6`** — `authbridgeconfig_controller.go` does not exist there, so `authbridge-config` stays as created (e.g. Helm-managed) and per-namespace realms are honored. Verified on a live cluster running `0.2.0-rc.6`: `client_credentials` for namespace clients returns HTTP 200. The breakage only appears when building the operator from `main` (which is what kagenti CI does via `20-build-operator.sh`, default ref `main`).

Introduced by: `cdbed3b` — "feat: operator-driven identity deployment" (2026-06-05).

## Proposed fix

Preserve an explicitly-set `KEYCLOAK_REALM` rather than clobbering it. Options:

1. **Don't overwrite a key that's already present / non-empty** in the existing CM (only fill in missing keys when reconciling an existing `authbridge-config`).
2. **Honor a per-namespace override** via a Namespace annotation/label (e.g. `kagenti.io/keycloak-realm`) that takes precedence over `Platform.KeycloakRealm`.
3. **Only manage CMs the reconciler owns** (skip CMs not labeled `app.kubernetes.io/managed-by: kagenti-operator`), so externally-provisioned config is left intact.

Option 1 or 2 restores per-namespace realm support without regressing the operator-driven defaulting added in `cdbed3b`.

## Cross-refs

- Interim mitigation in kagenti: the token-exchange E2E step was made non-fatal (kagenti/kagenti PR #1982) so `main` regains a green/red signal while this is fixed.

pdettori:
## Summary

`AuthbridgeConfigReconciler` unconditionally overwrote every key of an existing `authbridge-config` ConfigMap with platform-global defaults on each reconcile, silently discarding any explicitly-set per-namespace value such as `KEYCLOAK_REALM`. A workload in a namespace using a non-default realm had its realm reset to `kagenti` within one reconcile cycle, so the client-registration controller registered the client in the wrong realm and `client_credentials` grants returned `invalid_client` (401). Regression introduced in `cdbed3b`.

Fixes #433.

## Root cause

In the update path of `authbridgeconfig_controller.go`:

```go
for k, v := range desired {
    cm.Data[k] = v   // clobbers KEYCLOAK_REALM (and every other key)
}
```

## Fix

`authbridge-config` is a **namespace-defaults** source (labeled `kagenti.io/defaults=true`, consumed by `defaults_config_reconciler.go` / `agentruntime_controller.go`), and `cdbed3b` introduced it purely as operator-driven *defaulting*. Defaults should fill gaps, not override explicit values. On the update path the reconciler now only fills keys that are **missing or empty**, preserving any existing non-empty value:

```go
for k, v := range desired {
    if existing, ok := cm.Data[k]; ok && existing != "" {
        continue // preserve explicitly-set per-namespace value
    }
    cm.Data[k] = v
}
```

Unchanged: the **create** path (a fresh `kagenti-enabled` namespace still receives full operator-provided defaults) and the ownership/defaults **label stamping**. So the `cdbed3b` defaulting behavior is preserved for namespaces that have no config yet.

## Tests

The controller previously had **no test**. Adds `authbridgeconfig_controller_test.go` (plain `testing` + fake client, matching `clientregistration_controller_test.go`) covering:

| Case | Asserts |
|------|---------|
| Per-namespace realm preserved | existing `KEYCLOAK_REALM=tx-e2e` survives reconcile (the regression) |
| Create path | no CM → created with full platform defaults + labels |
| Missing/empty-key fill | present non-empty key kept; missing & empty keys filled from defaults |
| Label stamping | externally-created CM gets `managed-by` + `kagenti.io/defaults` |
| Skip | non-`kagenti-enabled` namespace → no CM created |

The first two/three cases were written failing first (clobber reproduced), then made green by the fix.

```
ok  github.com/kagenti/operator/internal/controller  (5/5 subtests pass)
```

`gofmt`, `go vet`, and `go build ./...` clean.

## Cross-refs

- Interim mitigation in kagenti made the token-exchange E2E step non-fatal (kagenti/kagenti #1982); this PR restores the underlying behavior so that suite can go green again.

_Assisted-By: Claude Code_

### DIFF
--- a/kagenti-operator/internal/controller/authbridgeconfig_controller.go
+++ b/kagenti-operator/internal/controller/authbridgeconfig_controller.go
@@ -111,7 +111,14 @@ func (r *AuthbridgeConfigReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if cm.Data == nil {
 		cm.Data = make(map[string]string)
 	}
+	// Fill in defaults only for keys that are missing or empty; preserve any
+	// explicitly-set per-namespace value (e.g. KEYCLOAK_REALM). This ConfigMap
+	// is a namespace-defaults source (kagenti.io/defaults=true), so defaults
+	// must not clobber a value provided by an operator/admin/CI. See issue #433.
 	for k, v := range desired {
+		if existing, ok := cm.Data[k]; ok && existing != "" {
+			continue
+		}
 		cm.Data[k] = v
 	}
 	if updateErr := r.Update(ctx, cm); updateErr != nil {
--- a/kagenti-operator/internal/controller/authbridgeconfig_controller_test.go
+++ b/kagenti-operator/internal/controller/authbridgeconfig_controller_test.go
@@ -1,0 +1,187 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+const authbridgeConfigTestNamespace = "tx-e2e"
+
+// authbridgeTestPlatform returns a platform config whose default realm is the
+// platform-global "kagenti" — i.e. the value that would clobber a per-namespace
+// override if the reconciler overwrote existing keys.
+func authbridgeTestPlatform() AuthbridgeConfigPlatform {
+	return AuthbridgeConfigPlatform{
+		KeycloakAdminSecretNamespace: "keycloak",
+		KeycloakRealm:                "kagenti",
+		KeycloakPublicURL:            "https://keycloak.example",
+	}
+}
+
+func kagentiEnabledNamespace(name string) *corev1.Namespace {
+	return &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   name,
+			Labels: map[string]string{labelKagentiEnabled: "true"},
+		},
+	}
+}
+
+func TestAuthbridgeConfigReconciler_Reconcile(t *testing.T) {
+	ctx := context.Background()
+	req := ctrl.Request{NamespacedName: types.NamespacedName{Name: authbridgeConfigTestNamespace}}
+
+	getCM := func(t *testing.T, c client.Client) *corev1.ConfigMap {
+		t.Helper()
+		cm := &corev1.ConfigMap{}
+		if err := c.Get(ctx, client.ObjectKey{Namespace: authbridgeConfigTestNamespace, Name: authbridgeConfigCM}, cm); err != nil {
+			t.Fatalf("get authbridge-config: %v", err)
+		}
+		return cm
+	}
+
+	cases := []struct {
+		name  string
+		objs  []client.Object
+		check func(t *testing.T, c client.Client)
+	}{
+		{
+			// Regression for #433: an explicitly-set per-namespace realm must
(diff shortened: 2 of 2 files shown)


esnible (approved):
## Summary

Clean, focused fix. The update path now fills only missing/empty keys instead of clobbering every key, so an explicitly-set per-namespace value (e.g. `KEYCLOAK_REALM=tx-e2e`) survives reconciliation while the create path still gets full operator-provided defaults. This matches Option 1 in #433 and correctly preserves the `cdbed3b` defaulting behavior for fresh namespaces.

Verified against `upstream/main`: the diff applies to the pre-fix clobbering loop; `LabelNamespaceDefaults == "kagenti.io/defaults"`, `authbridgeConfigCM`, and `clientRegistrationTestScheme` all exist and are used consistently; and `authbridge-config` is genuinely consumed as a namespace-defaults source (via `MatchingLabels{LabelNamespaceDefaults: "true"}`). Good first test for a controller that previously had none — the regression case is written to fail on the old behavior.

One behavior trade-off worth noting (non-blocking): once a key is non-empty, the operator can no longer push an updated platform default to it on the update path. That's the intended "defaults fill gaps, don't override" semantics per #433, so this is correct as designed.

**Author**: pdettori (MEMBER — maintainer)
**Areas reviewed**: Go (controller + test)
**Agent/IDE config (.claude/.vscode)**: none
**CI status**: passing (all 16 checks green, incl. E2E + Integration + Unit)

PR #438, opened 2026-06-16
Feat: add Kuadrant operand controller

Bobbins228:
## Summary

Adds a new operand controller to the kagenti-operator that reconciles a Kuadrant CR in `kuadrant-system`. This replaces the manual `kubectl apply` step in the setup scripts, providing idempotent creation and drift reconciliation.

## Changes

- `internal/controller/kuadrant_controller.go` — new `KuadrantReconciler` that ensures the `kuadrant-system` namespace and a bare Kuadrant CR exist, with `managed-by` label for drift detection. Includes a bootstrap `Runnable` to trigger initial reconcile (the `For()` watch alone cannot fire when no CR exists yet).
- `internal/controller/kuadrant_controller_test.go` — unit tests covering creation, label patching, no-op, and not-found scenarios
- `internal/kuadrant/types.go` — minimal Go types for `kuadrant.io/v1beta1` to avoid importing the full Kuadrant SDK
- `cmd/main.go` — conditional controller registration gated on `--enable-kuadrant` flag and `KuadrantCRDExists()` discovery check
- `charts/kagenti-operator/values.yaml` + `templates/manager/manager.yaml` — Helm values and template for the `kuadrant.enable` feature gate

## Test Plan

- [x] Deploy with Kuadrant CRD present and `--enable-kuadrant=true` — operator creates CR, Kuadrant operator reconciles to Ready
- [x] Verify `managed-by: kagenti-operator` label on created CR
- [x] Deploy without Kuadrant CRD and flag unset — operator starts cleanly, no Kuadrant activity
- [ ] Unit tests pass (`make test`)

### DIFF
--- a/charts/kagenti-operator/templates/rbac/role.yaml
+++ b/charts/kagenti-operator/templates/rbac/role.yaml
@@ -23,6 +23,7 @@ rules:
   - namespaces
   - services
   verbs:
+  - create
   - get
   - list
   - watch
@@ -142,6 +143,17 @@ rules:
   - statefulsets/finalizers
   verbs:
   - update
+- apiGroups:
+  - kuadrant.io
+  resources:
+  - kuadrants
+  verbs:
+  - create
+  - get
+  - list
+  - patch
+  - update
+  - watch
 - apiGroups:
   - discovery.k8s.io
   resources:
--- a/charts/kagenti-operator/templates/rbac/role.yaml
+++ b/charts/kagenti-operator/templates/rbac/role.yaml
@@ -20,12 +20,20 @@ rules:
   - ""
   resources:
   - endpoints
-  - namespaces
   - services
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
   verbs:
   - create
   - get
   - list
+  - patch
   - watch
 - apiGroups:
   - ""
--- a/kagenti-operator/internal/controller/kuadrant_controller.go
+++ b/kagenti-operator/internal/controller/kuadrant_controller.go
@@ -143,6 +143,8 @@ type kuadrantBootstrap struct {
 	reconciler *KuadrantReconciler
 }
 
+func (b *kuadrantBootstrap) NeedLeaderElection() bool { return true }
+
 func (b *kuadrantBootstrap) Start(ctx context.Context) error {
 	kuadrantLogger.Info("Bootstrap: triggering initial Kuadrant reconcile")
 	req := ctrl.Request{
--- a/kagenti-operator/internal/controller/kuadrant_controller.go
+++ b/kagenti-operator/internal/controller/kuadrant_controller.go
@@ -145,15 +145,28 @@ type kuadrantBootstrap struct {
 
 func (b *kuadrantBootstrap) Start(ctx context.Context) error {
 	kuadrantLogger.Info("Bootstrap: triggering initial Kuadrant reconcile")
-	_, err := b.reconciler.Reconcile(ctx, ctrl.Request{
+	req := ctrl.Request{
 		NamespacedName: types.NamespacedName{
 			Name:      kuadrantCRName,
 			Namespace: kuadrantNamespace,
 		},
-	})
-	if err != nil {
-		kuadrantLogger.Error(err, "Bootstrap reconcile failed — controller watch will retry on next event")
 	}
+
+	const maxAttempts = 5
+	for attempt := range maxAttempts {
+		_, err := b.reconciler.Reconcile(ctx, req)
+		if err == nil {
+			return nil
+		}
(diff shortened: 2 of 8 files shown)


charts/kagenti-operator/templates/rbac/role.yaml
```diff
@@ -23,6 +23,7 @@ rules:
   - namespaces
   - services
   verbs:
+  - create
```
pdettori, line 33: **suggestion (least-privilege):** `create` was added to the shared core rule that covers `endpoints`, `namespaces`, **and** `services`, so the operator ServiceAccount now also gets `create` on endpoints and services — but this controller only needs to create the `kuadrant-system` namespace.

A `make manifests` regeneration from your kubebuilder marker (`+kubebuilder:rbac:groups="",resources=namespaces,verbs=get;list;watch;create`) would split `namespaces` into its own rule with `create` and leave `endpoints`/`services` at `get;list;watch`. Recommend regenerating from the markers rather than hand-editing this rule.

kagenti-operator/internal/controller/kuadrant_controller.go
```diff
@@ -0,0 +1,196 @@
...
+func (b *kuadrantBootstrap) Start(ctx context.Context) error {
+	kuadrantLogger.Info("Bootstrap: triggering initial Kuadrant reconcile")
+	_, err := b.reconciler.Reconcile(ctx, ctrl.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      kuadrantCRName,
+			Namespace: kuadrantNamespace,
+		},
+	})
+	if err != nil {
+		kuadrantLogger.Error(err, "Bootstrap reconcile failed — controller watch will retry on next event")
```
pdettori: **suggestion (robustness):** the bootstrap swallows the reconcile error here with "controller watch will retry on next event" — but in the exact scenario this bootstrap exists for (no Kuadrant CR present yet), there is no CR to generate a watch event, so a transient failure means the CR is never created until the operator restarts. Consider a short bounded retry/backoff inside `Start()`. (Don't `return err` — that would crash the manager via the runnable group.)

pdettori (approved):
Solid operand controller. Feature-gated off by default (`--enable-kuadrant=false` + `kuadrant.enable: false` + CRD discovery guard), reuses the existing `LabelManagedBy`/`LabelManagedByValue` constants correctly, hand-rolls correct DeepCopy methods, and ships good unit coverage (create/patch/no-op/not-found). DCO passes.

The failing **E2E Tests** check is an unrelated SPIRE Helm-install timeout during test setup (30m test timeout at `helm install spire`), not caused by this gated-off feature — all other checks pass.

Approving with two non-blocking suggestions inline.

kevincogan (changes requested):
Good work addressing @pdettori's feedback. The RBAC split and bootstrap retry are both clean. One change needed before merge:

**`NeedLeaderElection()` missing on `kuadrantBootstrap` (kuadrant_controller.go)**

Every other bootstrap runnable in the repo (`OtelBootstrapRunnable`, `UIConfigBootstrapRunnable`) implements `NeedLeaderElection() bool` returning `true`. Without it, when leader election is enabled the bootstrap runs on every replica simultaneously. This causes redundant namespace/CR creation attempts and noisy logs. The `AlreadyExists` guards prevent data corruption, but this diverges from the established pattern and should be aligned.

One-liner fix:

```go
func (b *kuadrantBootstrap) NeedLeaderElection() bool { return true }
```

Two optional nits:

**1. Silent skip when flag enabled but CRD missing (main.go ~714-724)**
When `--enable-kuadrant=true` but `KuadrantCRDExists()` returns false, nothing is logged at `setupLog` level. Consider adding an `else` branch with `setupLog.Info("Kuadrant enabled but CRD not found; controller not started")` so operators can see why the controller did not start.

**2. Duplicate test case (kuadrant_controller_test.go)**
The last test "should handle not-found gracefully" uses an identical setup and assertions to the first test "should create namespace and Kuadrant CR when neither exists". Consider differentiating the scenario or removing the duplicate.

kevincogan (approved):
@Bobbins228 Nice work! Thanks for addressing the review feedback so quickly. The leader election fix, CRD-missing log, and the test dedup all look great. Appreciate you picking up the optional nits too.

Really solid contribution overall. The dual-gating approach (flag + CRD discovery) keeps this safe for clusters without Kuadrant, and the bootstrap runnable is a clean solution for bootstrapping a CR that doesn't exist yet. Commit history is well-structured and easy to review.

Approving, nice work!

Bobbins228:
Thanks @kevincogan I forgot to add a link in this PR but there is a companion [PR](https://github.com/kagenti/kagenti/pull/1988) kagenti side to remove Kuadrant CR creation from the ocp startup script if you have time to review that too.

PR #439, opened 2026-06-16
Fix: pin Helm binary to v4.2.0 to unblock E2E tests

Ygnas:
## Summary
The setup-helm action defaults to installing the latest Helm binary. Helm v4.2.1 introduced a regression in the wait/timeout behavior causing `helm install --wait --timeout 5m`
to hang indefinitely instead of failing after the specified timeout. This causes the SPIRE install step in E2E tests to consume the entire 30-minute test budget

Pin the Helm binary version to v4.2.0, which correctly respects the --timeout flag.

Fixes #


r3v5 (approved):
LGTM

akram (approved):

PR #441, opened 2026-06-17
Fix: add retry logic to metrics E2E curl pod to prevent flaky timeouts

r3v5:
## Summary
- The `curl-metrics` pod in the metrics E2E test had no connect timeout or retry logic
- A single TCP timeout from delayed KinD service routing caused the pod to fail, then the test polled for 5 minutes before giving up
- Added a retry loop (5 attempts, `--connect-timeout 15`, `--max-time 30`, 5s sleep between) so transient networking delays in KinD are handled gracefully

## Test plan
- [x] CI E2E tests pass (metrics test specifically)
- [x] Verify retry loop handles transient connection failures in KinD
- [x] Confirm no regression in other E2E tests

### DIFF
--- a/kagenti-operator/test/e2e/e2e_test.go
+++ b/kagenti-operator/test/e2e/e2e_test.go
@@ -214,7 +214,7 @@ var _ = Describe("Manager", Ordered, func() {
 								"name": "curl",
 								"image": "curlimages/curl:latest",
 								"command": ["/bin/sh", "-c"],
-								"args": ["curl -v -k -H 'Authorization: Bearer %s' https://%s.%s.svc.cluster.local:8443/metrics"],
+								"args": ["for i in 1 2 3 4 5; do curl -v -k --connect-timeout 15 --max-time 30 -H 'Authorization: Bearer %s' https://%s.%s.svc.cluster.local:8443/metrics && exit 0; echo \"Attempt $i failed, retrying in 5s...\"; sleep 5; done; exit 1"],
 								"securityContext": {
 									"allowPrivilegeEscalation": false,
 									"capabilities": {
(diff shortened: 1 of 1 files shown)


kevincogan (approved):
LGTM, clean fix for a real KinD flake. Retry logic is sound, timeouts are reasonable, and `%q` is a nice safety improvement over raw string interpolation.

Two optional nits:

**1. Consider `--fail` (`-f`) on curl (line 212)**
Without `-f`, curl exits 0 on HTTP 5xx, so the retry only covers transport-level failures. If the metrics server is reachable but returns 503 during startup, the pod exits `Succeeded` and the failure surfaces later at `getMetricsOutput()`. Adding `-f` would make the retry loop cover both transport and application-level transients.

**2. `Eventually` timeout could be tightened (line 261)**
With worst-case pod runtime at ~170s (5x30s + 4x5s), the 5-minute window is generous. Could reduce to ~3 minutes to fail faster on persistent failures.

Neither is blocking. Nice improvement to CI stability.

kagenti-operator/test/e2e/e2e_test.go
```diff
@@ -246,7 +258,7 @@ var _ = Describe("Manager", Ordered, func() {
 				g.Expect(err).NotTo(HaveOccurred())
 				g.Expect(output).To(Equal("Succeeded"), "curl pod in wrong status")
 			}
-			Eventually(verifyCurlUp, 5*time.Minute).Should(Succeed())
+			Eventually(verifyCurlUp, 3*time.Minute).Should(Succeed())
```
pdettori, line 261: **nit**: The 3-minute timeout leaves only ~10s buffer over the worst-case retry time (~170s). If the cluster is especially slow one iteration, a single flake could still trip the timeout. Consider 3m30s for a slightly more comfortable margin — though this is unlikely to matter in practice.

pdettori (approved):
Clean fix for a real KinD flake. The retry loop (5 attempts, connect-timeout 15s, max-time 30s, 5s sleep between retries) is well-structured. The `-f` flag covers HTTP-level errors, `%q` shell-escaping is safer than raw interpolation, and the tightened `Eventually` timeout is reasonable. No security concerns.

**Author**: r3v5 (CONTRIBUTOR — returning external)
**Areas reviewed**: Go (E2E test)
**Commits**: 3 commits, all signed-off ✓
**CI status**: all 16 checks passing

PR #444, opened 2026-06-18
Fix: add list/watch RBAC verbs for ServiceAccounts 

### LINKED ISSUE #426: Helm chart missing list/watch RBAC verbs for ServiceAccounts
## Problem

The Helm chart's ClusterRole (`kagenti-manager-role`) grants only `create`, `get`, `update` for ServiceAccounts:

```yaml
# charts/kagenti-operator/templates/rbac/role.yaml:59-62
- apiGroups: [""]
  resources: [serviceaccounts]
  verbs: [create, get, update]
```

Missing: **`list`** and **`watch`**.

## Impact

The webhook's `ensureServiceAccount` uses controller-runtime's cached client (`m.Client`), which backs every `Get` call with an informer. Informers require `list` + `watch` to sync their cache. When the informer tries to `list` ServiceAccounts, RBAC denies it, the informer never syncs, and the `Get` call hangs until timeout. This causes the webhook to return an error, and the API server rejects pod creation from Deployments/StatefulSets.

**Error in operator logs:**
```
serviceaccounts is forbidden: User "system:serviceaccount:kagenti-system:controller-manager"
cannot list resource "serviceaccounts" in API group "" at the cluster scope
```

## Why it wasn't caught

The e2e test suite uses `envtest` with a superuser kubeconfig — no RBAC restrictions apply. The issue only surfaces when deploying to a real cluster (e.g., Kind) via Helm.

## Fix

Add `list` and `watch` to the ServiceAccount RBAC rule in `charts/kagenti-operator/templates/rbac/role.yaml`:

```yaml
- apiGroups: [""]
  resources: [serviceaccounts]
  verbs: [create, get, list, update, watch]
```

Also update the corresponding kubebuilder RBAC marker in `pod_mutator.go` if one exists for ServiceAccounts.

## Discovered during

RHAIENG-4936 integration testing on Kind cluster with Helm-deployed operator.

gracesmith6504:
## Summary
The Helm chart's ClusterRole (`kagenti-manager-role`) only grants `create`, `get`, `update` for ServiceAccounts,
  missing `list` and `watch`. The webhook's `ensureServiceAccount` uses controller-runtime's cached client, which
  requires `list` + `watch` for its informer cache. Without them, the informer can't sync, `Get` calls hang until
  timeout, and pod creation is rejected.

  Changes:
  - Added `list` and `watch` verbs to ServiceAccount RBAC rule in Helm chart
  (`charts/kagenti-operator/templates/rbac/role.yaml`)
  - Same change in kustomize config (`kagenti-operator/config/rbac/role.yaml`)
  - Added kubebuilder RBAC marker for ServiceAccounts in `pod_mutator.go`

## Related issue(s)
https://github.com/kagenti/kagenti-operator/issues/426

Deploy the operator to a Kind cluster via Helm and verify that pods with SPIRE-enabled agents can be created without
  RBAC errors in the operator logs.

Fixes #426

### DIFF
--- a/charts/kagenti-operator/templates/rbac/role.yaml
+++ b/charts/kagenti-operator/templates/rbac/role.yaml
@@ -62,7 +62,9 @@ rules:
   verbs:
   - create
   - get
+  - list
   - update
+  - watch
 - apiGroups:
   - cert-manager.io
   resources:
--- a/kagenti-operator/config/rbac/role.yaml
+++ b/kagenti-operator/config/rbac/role.yaml
@@ -54,7 +54,9 @@ rules:
   verbs:
   - create
   - get
+  - list
   - update
+  - watch
 - apiGroups:
   - ""
   resources:
--- a/kagenti-operator/internal/webhook/injector/pod_mutator.go
+++ b/kagenti-operator/internal/webhook/injector/pod_mutator.go
@@ -96,6 +96,7 @@ func NewPodMutator(
 }
 
 // +kubebuilder:rbac:groups=core,resources=configmaps,verbs=create;get;list;patch;update;watch
+// +kubebuilder:rbac:groups=core,resources=serviceaccounts,verbs=create;get;list;update;watch
 
 // InjectAuthBridge evaluates the multi-layer precedence chain and conditionally injects sidecars.
 //


kevincogan (approved):
LGTM, nice fix! Clean and minimal. The root cause analysis in the issue is spot on. The cached client needs list + watch for the informer to sync, so without these verbs the ensureServiceAccount path just hangs on any real Helm deployment. Good catch keeping the Helm chart, kustomize config, and kubebuilder marker all in sync too. Approving pending all CI tests go green

r3v5 (approved):
lgtm

PR #448, opened 2026-06-18
Feat: AuthBridge TLS bridge — Phase 2 (operator coordination)

huang195:
## Summary

Phase 2 of the AuthBridge TLS bridge: the operator now provisions and wires the per-agent CA so a real agent's outbound HTTPS can be decrypted end-to-end by the AuthBridge forward proxy (proxy-sidecar / lite mode). Builds on the merged Phase 1 (`kagenti-extensions` #522), which added the `authlib/tlsbridge` package and the `tls_bridge` config schema.

The feature is **off by default**, gated on both a cluster feature gate and a per-workload opt-in.

## What's included

- **CRD**: `AgentRuntime.spec.tlsBridgeMode` (`disabled` | `enabled`, default `disabled`).
- **Feature gate**: `TLSBridge` (off by default).
- **Resolution**: `tlsBridgeMode` resolved CR > namespace > `disabled`, with a defensive enum fallback, mirroring the mtls resolution.
- **Validating webhook**: rejects `tlsBridgeMode=enabled` with any `authBridgeMode` other than `proxy-sidecar` / `lite` (the bridge lives in the Go forward proxy; it would be a no-op under envoy-sidecar / waypoint).
- **Per-agent CA reconciler**: ensures a namespace SelfSigned `Issuer` + an unconstrained CA `Certificate` (isCA, ECDSA P-256, cert-sign, **no** name constraints) whose Secret backs the bridge. Keyed on the **target workload name** (`spec.targetRef.name`), matching what the injecting webhook mounts — not the CR's `metadata.name`.
- **RBAC**: cert-manager `issuers` + `certificates` CRUD (regenerated config role + hand-synced chart role).
- **Mutator wiring**: renders the `tls_bridge` config block and hard-mounts the CA Secret (mode `0440`) into the sidecar (full keypair, to forge per-origin leaves) and every agent container (`ca.crt` + the common CA-bundle trust env vars), only when **both** the feature gate and the resolved per-workload mode are enabled. The hard mount blocks pod start until cert-manager issues the Secret, closing the controller-to-pod startup race.

Decoupled from SPIRE/mTLS; hard deps are proxy-sidecar/lite + cert-manager.

## Test Plan

- [x] `go build ./...`, `go vet ./...`
- [x] Unit tests (all packages except `test/e2e`): reconciler (workload-keyed Secret, gate/mode off paths), webhook reject, config-block render, end-to-end mutation (gate-on mounts + trust env; gate-off-but-enabled no-op)
- [x] `golangci-lint` clean on new lines
- [x] `make manifests` — no drift
- [ ] e2e on a live cluster (deferred to deploy; uses the Phase 1 `latest` image)

Assisted-By: Claude Code


kagenti-operator/internal/controller/tlsbridge_ca_controller.go
```diff
@@ -0,0 +1,102 @@
...
+	// Name the Secret after the TARGET WORKLOAD, not the AgentRuntime CR: the
+	// injecting webhook keys every per-agent resource (incl. this mount) off the
+	// workload name (resourceName == spec.targetRef.name), which can differ from
+	// the CR's metadata.name. Mismatch here => the webhook mounts a Secret that
+	// never exists => pod stuck Pending.
+	secretName := ar.Spec.TargetRef.Name + agentv1alpha1.TLSBridgeCASecretSuffix
+	cert := &cmv1.Certificate{ObjectMeta: metav1.ObjectMeta{Name: secretName, Namespace: ar.Namespace}}
+	if _, err := controllerutil.CreateOrUpdate(ctx, r.Client, cert, func() error {
+		cert.Spec = cmv1.CertificateSpec{
+			IsCA:        true,
```
cwiklik, line 81: **suggestion** (defense-in-depth, non-blocking) — the per-agent CA is **unconstrained** (no `NameConstraints`). The documented containment rationale (per-agent isolation + sidecar-only key + `0o444`) is sound, but cert-manager supports `nameConstraints` to scope the CA to the agent's actual egress hosts — so an exfiltrated sidecar key couldn't mint a leaf for *any* host. Worth considering as a future hardening.

kagenti-operator/internal/webhook/injector/volume_builder.go
```diff
@@ -17,10 +17,117 @@ limitations under the License.
...
+// Mode is 0444 on both. The sidecar runs non-root (Proxy.UID, gid != 0), so a
+// root-owned 0440 file would be unreadable without forcing a fixed pod fsGroup —
+// and a fixed fsGroup (we use 0) is rejected by OpenShift restricted-v2 SCC,
+// whose supplemental-group range is namespace-assigned. 0444 lets the non-root
+// sidecar read its key with no fsGroup at all; the key is still confined to the
+// sidecar container (never mounted into the agent), and ca.crt is a public cert.
+//
+// workloadName MUST equal the AgentRuntime's spec.targetRef.name (== crName in
+// the webhook) so the Secret name matches what the reconciler provisions.
+func applyTLSBridgeMounts(podSpec *corev1.PodSpec, workloadName string) {
```
cwiklik, line 78: **suggestion** — `applyTLSBridgeMounts` is guarded against double-injection (`volumeExists` / `addVolumeMountIfMissing`), but no test exercises the "already mounted" path. Since the mutating webhook can re-run on pod updates, a re-injection idempotency test would lock that guarantee.

cwiklik (approved):
High-quality, security-conscious operator feature (TLS bridge Phase 2), feature-gated **off by default**.

I verified the security crux directly: the per-agent CA **private key is confined to the sidecar** — the full keypair Secret mounts only into the proxy container, while agent containers get a `ca.crt`-only `Items` projection (`volume_builder.go`), so `tls.key` never reaches the agent. Mode `0o444` with no forced fsGroup keeps it OpenShift restricted-v2-SCC compatible; RBAC adds only namespace-scoped cert-manager `issuers`/`certificates` CRUD; no key material is logged.

Correctness also checks out: mode precedence (CR > namespace > default), webhook rejection of `enabled`+`envoy`/`waypoint` (tested), and the CA reconciler keys the Secret off the **workload name** (matching the webhook's mount) with owner references + idempotent `CreateOrUpdate`.

**Areas reviewed**: Go (controller/webhook/mutator/api), CRD/RBAC, Helm, Docs, Security, Tests
**Commits**: 18, all signed-off (DCO ✓)
**CI**: all green (E2E, Integration, Unit, CodeQL, Trivy, Lint, Helm)

Approving — two non-blocking defense-in-depth/coverage notes inline.

<sub>Reviewed with Claude Code (`/github:pr-review`).</sub>

PR #449, opened 2026-06-18
Feat: Add token-broker service for OAuth session and token management

### LINKED ISSUE #455: feature: Token Broker Service
### Feature Description

The token-broker service provides centralized OAuth token acquisition for AI agents via PKCE-based authorization flows, token caching with JWT expiry parsing, and session management with long-polling coordination.

### Proposed Solution

A centralized service for maintaining user sessions and channel to users to redirect them to OAuth systems. 
The service serve also as the OAuth Client  - future enhancement should allow external IdP (KeyClock) to serve as the OAuth 2client instead.

### Want to contribute?

- [x] I would like to work on this issue.

### Additional Context

_No response_

davidhadas:
## Summary
The token-broker service provides centralized OAuth token acquisition for AI agents via PKCE-based authorization flows, token caching with JWT expiry parsing, and session management with long-polling coordination.

Fixes:  https://github.com/kagenti/kagenti-operator/issues/455 

## Related issue(s)
https://github.com/kagenti/kagenti/issues/1435


rhuss:
Thanks for the work on this, David. I don't want to block your research progress, but a few observations from the productization side might be still ok:

**Completeness:** The token-broker cannot complete an OAuth flow end-to-end without the Backend service (Phase 3) that handles the user-facing browser interaction. JWT signature validation is also explicitly skipped. As it stands, this is a Phase 1 foundation, not a shippable component.

**Keycloak overlap:** The core functionality (authorization code + PKCE, token caching, JWT expiry parsing, session management) overlaps significantly with what Keycloak provides natively through identity brokering. The implementation also reimplements standard Go library functionality (`golang.org/x/oauth2`, `golang-jwt/jwt`) rather than using them, which increases the maintenance and CVE surface without clear benefit. The legitimate gap (bridging to external providers not federated with Keycloak) can potentially be addressed through Keycloak identity brokering configuration or CIBA, both of which are production-hardened paths that don't require a new service.

**Repo integration:** The token-broker is architecturally independent (separate binary, separate Deployment, zero imports from operator code), but it shares the operator's `go.mod`. This means downstream consumers who don't ship the token-broker still inherit its dependencies in SBOM/vulnerability scans and need to carry deletion patches that require `go mod tidy` after every upstream sync. Since the code is fully decoupled, moving it to its own Go module (separate `go.mod`, either as a subdirectory module or a separate repo) would make it cleanly excludable without per-sync patching.

For a long term solution, maybe consider extracting the token-broker into its own module/repo and evaluating whether Keycloak identity brokering or CIBA could address the external-provider token acquisition gap before building a custom service. Happy to discuss the Keycloak/CIBA angle further (see also the [discussion on #1435](https://github.com/kagenti/kagenti/issues/1435#issuecomment-4687914468)).

davidhadas:
Thanks for the review. 
A Backend service  - see https://github.com/kagenti/kagenti/pull/1583
JWT signature validation  - committed

> As it stands, this is a Phase 1 foundation, not a shippable component.

Agreed. We are laying down the HITL foundation. More work is definitely needed, but we are at a phase where we can finally see the landscape and discuss the design decisions, controls, configs, and user experience.

Does this mean we should hold off on merging this PR once it passes review? Or can we merge it post-review with the explicit note that we still have work before it is production-ready?

Repo integration: Understood. So, keep it in this repo but ensure it uses its own isolated go.mod.

Relationship with Keycloak: Let’s keep that specific discussion over to the Epic issue https://github.com/kagenti/kagenti/issues/1435 as it is part of the big picture we need to discuss. 

It should be noted here though that token broker service should have a later PR to include integration with KeyCloak supporting use cases in which Kagneti is deployed with KeyCloak and need to offer OAuth 2 against federated systems.

rhuss:
Thanks David. To be clear: please don't feel blocked by me on merging this. My comments were architectural observations, not merge blockers.

I've come to better appreciate that kagenti's upstream is designed for research incubation and rapid iteration, without stability guarantees that would slow contributors down. That's a valid and valuable mode of operation. The productization constraints I've been raising come from a different set of goals, and it's not fair to apply those as a gate on upstream research work.

Go ahead and merge when it passes review. The foundation you're laying here is solid exploration of the HITL problem space. Separate `go.mod` would still be appreciated for cleaner downstream consumption, but that's a housekeeping ask, not a blocker.

davidhadas:
@rhuss , 
The token broker service was moved to its own Go module.

token-broker/internal/cache/token_cache.go
```diff
@@ -0,0 +1,170 @@
...
+
+	if sessionTokens, ok := tc.tokens[sessionKey]; ok {
+		return len(sessionTokens)
+	}
+	return 0
+}
+
+// CleanupExpiredTokens removes all expired tokens from the cache.
+// This can be called periodically to free memory.
+func (tc *TokenCache) CleanupExpiredTokens() int {
```
akram, line 143: `CleanupExpiredTokens()` is defined but never wired to a ticker in main.go. For long-running instances, expired tokens will accumulate in memory. Suggestion: add a periodic cleanup
  goroutine, or clean up lazily on SetToken when the session map exceeds a threshold.

token-broker/internal/core/interfaces.go
```diff
@@ -0,0 +1,176 @@
...
+	// SetToken stores a token for a session and resource server.
+	SetToken(sessionKey, resourceURL, token string) error
+
+	// DeleteToken removes a token from the cache.
+	DeleteToken(sessionKey, resourceURL string)
+}
+
+// OAuthDiscoverer performs OAuth discovery against a resource server.
+// Updated to support the new architecture where Token Broker is the OAuth client.
+type OAuthDiscoverer interface {
```
akram, line 53: the `OAuthDiscoverer` interface is tightly coupled to the OAuth authorization code + PKCE flow.
  If we want to support alternative token acquisition mechanisms (e.g., Keycloak identity brokering, CIBA, or custom IdP flows), it might be worth defining a higher-level TokenAcquirer
  interface that abstracts the grant type entirely, with PKCE as one implementation.

token-broker/internal/api/handlers.go
```diff
@@ -0,0 +1,521 @@
...
+// writeError writes a standardized error response (flat format for AuthBridge API).
+func writeError(w http.ResponseWriter, status int, code, message string) {
+	writeJSON(w, status, ErrorResponse{
+		Code:    code,
+		Message: message,
+	})
+}
+
+// writeBackendError writes a standardized error response (nested format for Backend API).
+func writeBackendError(w http.ResponseWriter, status int, code, message string) {
```
akram, line 139: The API uses two different error response formats: `writeError (flat, for AuthBridge)` and `writeBackendError (nested, for Backend)`. This makes client integration harder — consumers need
  to know which endpoint returns which format. Consider standardizing on a single error envelope.

token-broker/deploy/02-deployment.yaml
```diff
@@ -0,0 +1,115 @@
...
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: token-broker
+  namespace: kagenti-system
+  labels:
+    app: token-broker
+spec:
+  replicas: 1 # In-memory sessions/token cache: do NOT scale >1 without shared state.
```
akram, line 17: Already discussed in the Slack thread — the `replicas: 1` constraint with in-memory state is a known limitation. Just noting that the security posture is good (non-root, read-only
  filesystem, capabilities dropped).

akram (changes requested):

akram:
## Keycloak overlap question

Roland's [observation](https://github.com/kagenti/kagenti-operator/pull/449#issuecomment-4687914468) about the overlap with Keycloak identity brokering is worth exploring further. The core of what the token-broker implements — authorization code + PKCE, token caching, JWT expiry parsing, session management — maps closely to what Keycloak provides natively. CIBA (Client-Initiated Backchannel Authentication) is also an interesting alternative for the HITL use case: the backend initiates the auth request, the user consents via a separate channel, no browser redirect needed.

## Provider-agnostic token brokering interface

That said, tying the HITL flow exclusively to Keycloak would create a hard dependency on a specific IdP. Since we're actively working on decoupling from Keycloak on the AuthBridge side, it raises a broader design question: could we define a **provider-agnostic interface for token brokering**, where Keycloak is one possible backend but not the only one?

Concretely, the current `OAuthDiscoverer` interface in `interfaces.go` is tightly coupled to the authorization code + PKCE flow. A higher-level `TokenAcquirer` abstraction could support multiple grant types — PKCE, CIBA, identity brokering, or custom IdP flows — so the HITL mechanism works regardless of the underlying IdP (Keycloak, external OIDC provider, or something else entirely).

davidhadas:
@akram  - all good points. Thank you for the review. I addressed one by one. 

Let's record the overall solution directions discussions in the epic https://github.com/kagenti/kagenti/issues/1435 - I think some of your comments related to Keycloak overlap question  were already discussed there and maybe you can comment when it was not.

token-broker/internal/cache/token_cache.go
davidhadas: let me fix that.

token-broker/internal/core/interfaces.go
davidhadas: Good point on the interface design. The current OAuthDiscoverer is intentionally scoped to the authorization code + PKCE flow because that is the only grant type in scope for this phase. Introducing a higher-level TokenAcquirer abstraction that covers PKCE, CIBA, identity brokering, etc. makes sense as a follow-up when a second grant type is actually needed — adding it now seems premature.

token-broker/internal/api/handlers.go
davidhadas: The two formats are intentional and serve two different consumer contracts.

- The Backend API uses the nested {"error": {"code": ..., "message": ...}} format, which aligns with the conventional OAuth 2.0 / RFC 7807 error envelope that external service consumers expect. This is the right format for that surface and should not change.
- The AuthBridge API uses a flat {"code": ..., "message": ...} format because AuthBridge is an internal sidecar component with its own established error-parsing contract. Changing it here would require a coordinated change in the AuthBridge plugin code, which is out of scope for this PR. 

Aligning the two formats is a valid future improvement, but it carries the risk of breaking the AuthBridge integration if not done carefully and belongs in a dedicated PR targeting that side of the interface.

token-broker/deploy/02-deployment.yaml
davidhadas: Agreed. Scaling beyond 1 replica and moving to persistent state is left for a later revision. 
The current focus is on getting the functionality right while keeping non-functional work to the minimum necessary — without compromising on security.

davidhadas:
service error codes are wrong
```
accessToken, err := h.broker.AcquireToken(ctx, sessionKey, userID, resourceURL)
if err != nil {
    if err.Error() == "timeout waiting for OAuth completion" {   // 408
    if err.Error() == "session expired" || err.Error() == "session ended" {  // 401
    writeError(w, http.StatusServiceUnavailable, "oauth_failed", ...)  // 503 default
}
```

review and improve all error handling and replaced with:

```
accessToken, err := h.broker.AcquireToken(ctx, sessionKey, userID, resourceURL)
if err != nil {
h.logger.Error("Token acquisition failed",
	"session_key", sessionKey,
	"user_id", userID,
	"resource_url", resourceURL,
	"error", err)

// Classify the error with errors.Is so that wrapping (AcquireToken wraps
// every performOAuthFlow error) does not collapse a specific, recoverable
// condition into a generic 5xx. A 5xx makes agents/sidecars mark this route
// dead and stop retrying, so it is reserved for genuinely retryable
// server/upstream failures — never session/client state.
switch {
case errors.Is(err, core.ErrOAuthTimeout):
	// User did not finish the OAuth flow in time — transient, retryable.
	writeError(w, http.StatusRequestTimeout, "timeout", "OAuth flow did not complete in time")
case errors.Is(err, core.ErrSessionEnded), errors.Is(err, core.ErrSessionExpired):
	// Session gone — permanent for this session; client must re-establish.
	writeError(w, http.StatusUnauthorized, "session_expired", "Session ended or expired")
case errors.Is(err, core.ErrOAuthUpstream):
	// Discovery / token exchange failed upstream — transient, retryable.
	writeError(w, http.StatusBadGateway, "oauth_upstream", "Upstream OAuth provider error")
default:
	// Unknown: do not signal a permanently-dead route. Retryable server error.
	writeError(w, http.StatusInternalServerError, "internal_error", "Failed to obtain token")
}
```

token-broker/internal/api/handlers.go
akram: ok. I will create follow up issue for that as well.

token-broker/cmd/main.go
```diff
@@ -0,0 +1,372 @@
...
+			case <-ticker.C:
+				removed := tokenCache.CleanupExpiredTokens()
+				if removed > 0 {
+					logger.Info("Periodic token cache cleanup", "removed", removed)
+				}
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
```
akram, line 357: lgtm

akram (approved):

PR #451, opened 2026-06-19
Feat: Support Sandbox workloads in AuthBridge owner resolution + ConfigMap GC

huang195:
## Summary

Make the AuthBridge webhook + operator handle **`Sandbox`** (`agents.x-k8s.io`) workloads with the same robustness as Deployment/StatefulSet, so the kagenti backend can attach an `AgentRuntime` (`mtlsMode` / `authBridgeMode` / `tlsBridgeMode`, including the TLS bridge) to sandbox-workload agents.

Three additive, case-only changes:

- **`ResolveOwner` Sandbox case** (`internal/workload/owner.go`) — resolve a Sandbox pod's owner as `{Name: <sandbox>, Kind: "Sandbox"}` (mirrors the existing `IsPodOwnedBy`), so AuthBridge override matching is by exact kind+name instead of the pod-name fallback. Closes the edge case where two AgentRuntimes share a `targetRef.name` across different kinds.
- **`buildOwnerReference` Sandbox branch** (`internal/webhook/injector/pod_mutator.go`) — set an ownerRef to the Sandbox so the per-agent `authbridge-config-<name>` ConfigMap is garbage-collected with the Sandbox (parity with Deployment/StatefulSet; previously orphaned on delete).
- **`getPodSpec` nil-guard** (`internal/controller/agentruntime_controller.go`) — the Sandbox pod-template accessor left `getPodSpec` unset (dead code today); return nil from a real func so a future caller hits a nil-check rather than a nil-function-call panic.

The CA reconciler and validating webhook are already kind-agnostic, and the AgentRuntime controller already supports Sandbox targetRefs (two-phase scale-restart), so this PR is purely robustness + GC. **No change to Deployment/StatefulSet/Job paths.**

## Test plan

- `go test ./internal/workload/... ./internal/webhook/injector/...` — including new `TestResolveOwner_Sandbox` and `TestEnsurePerAgentConfigMap_OwnerReference_SetFromSandbox`.
- Verified end-to-end on Kind: a sandbox `weather-agent` with `tlsBridgeMode: enabled` gets the per-agent CA minted, the two-volume CA mount + `tls_bridge` config injected, forged-leaf TLS interception (`issuer=CN=authbridge-tls-bridge-ca-weather-agent`), and decrypted request payloads (LLM `/v1/chat/completions`, MCP `/mcp`) on the session API.

## Notes

Pairs with the kagenti backend change that creates the Sandbox-targetRef AgentRuntime (separate PR in `kagenti/kagenti`). This operator change is optional hardening — the released operator already handles sandbox via the pod-name fallback; this makes it exact and adds ConfigMap GC.

### DIFF
--- a/kagenti-operator/internal/controller/agentruntime_controller.go
+++ b/kagenti-operator/internal/controller/agentruntime_controller.go
@@ -1331,6 +1331,12 @@ func newRuntimePodTemplateAccessor(kind string) (*runtimePodTemplateAccessor, bo
 				u := o.(*unstructured.Unstructured)
 				_ = unstructured.SetNestedStringMap(u.Object, a, "spec", "podTemplate", "metadata", "annotations")
 			},
+			// getPodSpec is not implemented for Sandbox: extracting a typed
+			// *corev1.PodSpec out of the unstructured podTemplate isn't needed
+			// today (the only former caller, OCI skill mounting, was removed).
+			// Return nil rather than leaving the func nil so a future caller hits
+			// a nil-check, not a nil-function-call panic.
+			getPodSpec: func(client.Object) *corev1.PodSpec { return nil },
 		}, true
 	default:
 		return nil, false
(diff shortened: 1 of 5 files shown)


kevincogan (approved):
LGTM. Approve. The `ResolveOwner` Sandbox case, ConfigMap GC via `buildOwnerReference`, nil-guard, and tests all look correct.

One observation for a follow-up: the per-agent ConfigMap naming (`authbridge-config-<name>`) and owner lookup are keyed by workload name only, not name+kind. If two workload kinds share a name in the same namespace, they'd collide. This is a pre-existing design assumption across all three supported kinds (Deployment, StatefulSet, Sandbox), not specific to this PR.

PR #456, opened 2026-06-22
Fix(helm): align ClusterRole with controller-gen and add AuthBridge SCC

Bobbins228:
## Summary

The Helm chart's ClusterRole has drifted from the controller-gen output (`config/rbac/role.yaml`), granting excess permissions. The RBAC markers in the Go source (`+kubebuilder:rbac`) are the single source of truth — `make manifests` generates `config/rbac/role.yaml` from them, but the Helm chart's copy is manually maintained with no sync step (unlike CRDs which have `sync-chart-crds`). This PR realigns them.

Additionally, the AuthBridge SecurityContextConstraints — required for OpenShift deployments — exists in the kustomize path (`config/security/kagenti-authbridge-scc.yaml`) but is missing from the Helm chart.

### ClusterRole alignment (`role.yaml`)

Rewrite the Helm chart ClusterRole to exactly match `config/rbac/role.yaml` (verified via `make manifests` + diff). Specific fixes:

- **Remove unused `endpoints` rule** — no `+kubebuilder:rbac` marker exists for core endpoints in the codebase
- **Remove `update` from namespaces** — controller-gen only grants `create/get/list/patch/watch`
- **Remove `patch`/`watch` from services** — controller-gen only grants `create/get/list/update/watch`
- **Remove `list` from `operator.openshift.io/networks`** — controller-gen only grants `get`
- **Remove `delete` from roles** — only `rolebindings` needs `delete`
- **Merge `configmaps` + `secrets`** into one rule (same verbs, matches controller-gen)
- **Merge `serviceaccounts` + `services`** into one rule (same verbs, matches controller-gen)
- **Reorder rules** to match controller-gen alphabetical output

Also regenerates `config/rbac/role.yaml` via `make manifests` to pick up controller-gen's latest normalisation (no permission changes, just grouping/ordering).

### AuthBridge SCC template (`authbridge-scc.yaml`) — new file

Adds three resources for OpenShift, mirroring `config/security/kagenti-authbridge-scc.yaml`:

1. **SecurityContextConstraints** (`kagenti-authbridge`) — least-privilege SCC allowing CSI volumes for SPIRE workload identity
2. **ClusterRole** (`system:openshift:scc:kagenti-authbridge`) — grants `use` verb scoped to `resourceNames: ["kagenti-authbridge"]`
3. **ClusterRoleBinding** — binds the operator ServiceAccount to the ClusterRole

All three are guarded by `{{- if and .Values.rbac.enable (.Capabilities.APIVersions.Has "security.openshift.io/v1") }}` so they only render on OpenShift.

## Test Plan

- [ ] `helm lint charts/kagenti-operator/` passes
- [ ] `helm template` without `--api-versions security.openshift.io/v1` does **not** render `authbridge-scc.yaml`
- [ ] `helm template --api-versions security.openshift.io/v1` renders SCC, ClusterRole, and ClusterRoleBinding
- [ ] `diff <(sed -n '/^rules:/,$p' config/rbac/role.yaml) <(sed -n '/^rules:/,$p' charts/.../role.yaml)` produces no output
- [ ] Existing Helm-based deployments continue working (permissions only removed, not added)

### DIFF
--- a/charts/kagenti-operator/templates/rbac/role.yaml
+++ b/charts/kagenti-operator/templates/rbac/role.yaml
@@ -9,6 +9,7 @@ rules:
   - ""
   resources:
   - configmaps
+  - secrets
   verbs:
   - create
   - delete
@@ -23,28 +24,16 @@ rules:
   - endpoints
   verbs:
   - get
-  - list
-  - watch
 - apiGroups:
   - ""
   resources:
   - namespaces
-  - services
   verbs:
   - create
   - get
   - list
   - patch
-  - update
   - watch
-- apiGroups:
-  - ""
-  - events.k8s.io
-  resources:
-  - events
-  verbs:
-  - create
-  - patch
 - apiGroups:
   - ""
   resources:
@@ -58,46 +47,22 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - secrets
+  - serviceaccounts
+  - services
   verbs:
   - create
-  - delete
   - get
   - list
-  - patch
   - update
   - watch
 - apiGroups:
   - ""
+  - events.k8s.io
   resources:
-  - serviceaccounts
+  - events
   verbs:
   - create
-  - get
-  - list
-  - update
-  - watch
-- apiGroups:
-  - cert-manager.io
-  resources:
-  - certificates
-  - issuers
-  verbs:
-  - create
-  - delete
-  - get
-  - list
   - patch
-  - update
-  - watch
-- apiGroups:
-  - cert-manager.io
-  resources:
-  - clusterissuers
-  verbs:
-  - get
-  - list
-  - watch
 - apiGroups:
   - agent.kagenti.dev
   resources:
@@ -165,6 +130,13 @@ rules:
 - apiGroups:
   - apps
   resources:
+  - deployments/finalizers
+  - statefulsets/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - apps
+  resources:
   - statefulsets
   verbs:
   - create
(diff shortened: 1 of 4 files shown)


charts/kagenti-operator/templates/rbac/role.yaml
```diff
@@ -17,34 +18,16 @@ rules:
   - patch
   - update
   - watch
-- apiGroups:
-  - ""
-  resources:
-  - endpoints
```
ChristianZaccaria, line 23: We may want to keep `endpoints` with at least the `get` role.

I found there's a function that performs a `Get` operation on `corev1.Endpoints`, found here: https://github.com/kagenti/kagenti-operator/blob/main/kagenti-operator/internal/controller/agentcard_networkpolicy_controller.go#L346-L351 

Without this this call will fail and the network policies will deny all egress, which means agent pods won't be able to reach the K8s API server.

ChristianZaccaria (changes requested):
Great work Mark! - I left one requested change. After that, lgtm!

pdettori (commented):
Reviewed the RBAC alignment and AuthBridge SCC addition. The ClusterRole changes correctly match `controller-gen` output, the SCC template is well-structured (proper capability guard, minimal privileges, operator SA delegation pattern), and CI is all green.

The only blocking issue (missing `endpoints` rule) has already been identified by @ChristianZaccaria — the controller performs a `Get` on `corev1.Endpoints` at `agentcard_networkpolicy_controller.go:346` but lacks a corresponding `+kubebuilder:rbac` marker, so controller-gen legitimately excludes it. The fix should add the marker and re-run `make manifests` to keep the Helm chart and kustomize output in sync.

charts/kagenti-operator/templates/rbac/role.yaml
Bobbins228: Great catch Christian, updated the commit

Bobbins228:
@pdettori, @ChristianZaccaria updated the PR thanks for the reviews

ChristianZaccaria (approved):
/lgtm /approve

r3v5:
Related to #460 right?

r3v5 (approved):
lgtm

pdettori (approved):
LGTM

PR #457, opened 2026-06-22
Fix(e2e): remove hardcoded runAsUser from test fixtures

Bobbins228:
## Summary

All E2E test fixtures hardcode `runAsUser: 1000` or `runAsUser: 65534` in pod security contexts. This is unnecessarily prescriptive — the actual security intent is `runAsNonRoot: true`, which is already set on every fixture. The hardcoded UIDs break on any platform with UID range enforcement (e.g. OpenShift SCCs assign UIDs from the namespace's allocated range, and a fixed UID outside that range causes pod admission failures).

Removing them has no effect on Kind, where any non-root UID is accepted.

## Changes

- Remove `runAsUser: 1000` from `fixtures.go` (17 occurrences), `e2e_test.go` (1 occurrence)
- Remove `runAsUser: 65534` from `sharedtrust_test.go` (2 occurrences)
- Preserve `runAsNonRoot: true` and `seccompProfile` on all pod specs (21 total, unchanged)

## Test Plan

- [ ] \`go build ./test/e2e/\` compiles cleanly
- [ ] \`grep -r runAsUser test/e2e/\` returns no results
- [ ] \`grep -c runAsNonRoot test/e2e/fixtures.go test/e2e/e2e_test.go test/e2e/sharedtrust_test.go\` confirms all security contexts preserved
- [ ] All 41 E2E specs pass on Kind (no regression)

### DIFF
--- a/kagenti-operator/test/e2e/e2e_test.go
+++ b/kagenti-operator/test/e2e/e2e_test.go
@@ -233,7 +233,6 @@ var _ = Describe("Manager", Ordered, func() {
 										"drop": ["ALL"]
 									},
 									"runAsNonRoot": true,
-									"runAsUser": 1000,
 									"seccompProfile": {
 										"type": "RuntimeDefault"
 									}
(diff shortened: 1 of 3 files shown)


Bobbins228:
Collapsing into PR #458 — the runAsUser removal only makes sense alongside the configurable images/env vars.

PR #401, opened 2026-06-03
Feat: [spec] mTLS transport security for agent communication (003)

varshaprasad96:
## Summary

- Adds spec-driven development artifacts for mTLS transport security (spec 003)
- Covers two communication paths: controller-to-agent and agent-to-agent mTLS
- SPIRE is the sole certificate provider; Istio is explicitly out of scope
- mTLS defaults to `permissive` (enabled by default, opt-out with `mTLSMode: disabled`)
- Authbridge mTLS is already implemented — remaining work is primarily operator-side

## Artifacts

| File | Purpose |
|------|---------|
| `specs/003-mtls-transport-security/spec.md` | Feature specification with user stories, requirements, acceptance scenarios |
| `specs/003-mtls-transport-security/plan.md` | Implementation plan with constitution check and project structure |
| `specs/003-mtls-transport-security/research.md` | Research decisions (config schema, SpiffeFetcher, MTLSReady condition) |
| `specs/003-mtls-transport-security/data-model.md` | Entity changes, condition states, state transitions |
| `specs/003-mtls-transport-security/tasks.md` | 24 implementation tasks across 8 phases |
| `specs/003-mtls-transport-security/contracts/` | Operator-to-authbridge mTLS config contract |
| `specs/003-mtls-transport-security/brainstorm.md` | Design exploration and clarification answers |

## Key Decisions

- **SPIRE only** — no Istio dependency
- **mTLS enabled by default** — `permissive` mode, operators opt out with `mTLSMode: disabled`
- **Fail clearly** — `MTLSReady=False` condition when SPIRE is unavailable
- **Controller uses go-spiffe SDK** directly (SpiffeFetcher as default)
- **spiffe-helper** for data-plane certificates (file-based, sidecar-agnostic)
- **JWS signing deprecated** — flags default to false with warnings

## References

- ADR: ODH-ADR-AgentOps-0002
- Jira: RHAIENG-4944
- Parent: RHAISTRAT-1599

## Test plan

- [ ] Team reviews spec, plan, and task breakdown
- [ ] Validate assumptions about authbridge mTLS state on main
- [ ] Confirm ConfigMap schema matches authbridge expectations
- [ ] Agree on MVP scope (US1: agent-to-agent mTLS by default)

Signed-off-by: Varsha Prasad Narsing <varshaprasad96@gmail.com>
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #461, opened 2026-06-22
Feat: add MTLSReady condition to AgentRuntime status

varshaprasad96:
## Summary

- Add `MTLSReady` condition logic to the AgentRuntime reconciler (T006 from mTLS spec)
- Controller evaluates SPIRE availability on each reconcile by checking for `spire-agent-socket` or `svid-output` volumes in the workload's pod template
- Emits Warning Event with actionable guidance when SPIRE is missing but mTLS is enabled

## Condition States

| Reason | Status | When |
|--------|--------|------|
| `SPIREAvailable` | True | SPIRE volumes detected in workload pod template |
| `SPIREUnavailable` | False | mTLS mode is permissive/strict but no SPIRE volumes found |
| `MTLSDisabled` | True | mTLSMode explicitly set to `disabled` |
| `SPIREAssumed` | True | Sandbox workloads (webhook handles injection at pod CREATE) |

## Key Design Decisions

- `MTLSReady=False` does **NOT** block `Ready=True` — it's informational for fleet-wide mTLS rollout tracking
- SPIRE detection checks pod template volumes, not running pod state — this is the right layer since the webhook injects volumes at admission time
- Sandbox workloads assume SPIRE is available since typed PodSpec is not accessible from unstructured objects

## Test plan

- [x] All unit tests pass (`make test`)
- [ ] Deploy agent with SPIRE → verify `MTLSReady=True/SPIREAvailable`
- [ ] Deploy agent without SPIRE → verify `MTLSReady=False/SPIREUnavailable` + Warning Event
- [ ] Deploy agent with `mTLSMode: disabled` → verify `MTLSReady=True/MTLSDisabled`

Jira: RHAIENG-4928
Spec: specs/003-mtls-transport-security (PR #401)

### DIFF
--- a/kagenti-operator/config/rbac/role.yaml
+++ b/kagenti-operator/config/rbac/role.yaml
@@ -41,15 +41,6 @@ rules:
   - ""
   resources:
   - serviceaccounts
-  verbs:
-  - create
-  - get
-  - list
-  - update
-  - watch
-- apiGroups:
-  - ""
-  resources:
   - services
   verbs:
   - create
@@ -169,17 +160,6 @@ rules:
   - list
   - watch
 - apiGroups:
-  - k8s.keycloak.org
-  resources:
-  - keycloakrealmimports
-  - keycloaks
-  verbs:
-  - create
-  - get
-  - list
-  - patch
-  - update
-- apiGroups:
   - datasciencecluster.opendatahub.io
   resources:
   - datascienceclusters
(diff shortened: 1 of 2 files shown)


pdettori (approved):
Clean implementation of the informational `MTLSReady` condition (T006). Code follows existing controller patterns consistently — accessor, condition API, event emission with nil-check guard.

**Verified:**
- `MTLSReady=False` does NOT block `Ready=True` ✓
- Sandbox path returns `SPIREAssumed` via nil-podSpec check (not nil-function-pointer) ✓
- RBAC diff is controller-gen consolidation — no permission change (serviceaccounts and services already had identical verbs) ✓
- Warning Event emits actionable guidance ✓
- All 16 CI checks passing ✓

PR #466, opened 2026-06-23
Refactor: remove IdentitySpec and SPIFFEIdentity from AgentRuntime CRD

Ygnas:
## Summary
Remove the Identity field from AgentRuntimeSpec along with the IdentitySpec and SPIFFEIdentity struct definitions. These types were defined in the API surface and generated into deepcopy/CRD schemas but were never consumed by any controller or webhook logic at runtime, making them dead CRD surface area.

- Remove Identity field, IdentitySpec, and SPIFFEIdentity from agentruntime_types.go
- Regenerate deepcopy and CRD manifests via controller-gen
- Sync Helm chart CRDs
- Update controller and e2e tests to remove identity references and add mtlsMode: strict as the override differentiator
- Update sample CRs to remove identity blocks
- Correct authbridge-webhook.md: remove fictional Layer 3 / ReadAgentRuntimeOverrides section — the webhook never reads AgentRuntime CR spec fields; mode resolution uses the namespace authbridge-runtime-config ConfigMap exclusively
- Fix merge code path diagram to match actual ResolveConfig signature (2 args, not 3)

Fixes #


pdettori (approved):
## Summary

APPROVE. Verified the core claim against source on both `main` and this branch, not just the diff:

- **Dead-code removal is safe.** On `main`, `Identity`/`IdentitySpec`/`SPIFFEIdentity` appear only in `agentruntime_types.go` (definitions) and `zz_generated.deepcopy.go` (generated) — **zero** consumers in any controller or webhook code path. Removal is corroborated by the green Build / Unit / Integration / E2E checks.
- **The doc corrections are accurate, not hand-wavy.** `ReadAgentRuntimeOverrides()` — described by the old `authbridge-webhook.md` as the webhook's "Layer 3" reader of `spec.identity` — exists nowhere in the repo (main or branch), so the old "Layer 3 / spec.identity" section was genuinely fictional. The rewrite matches reality: `ResolveConfig(platform, ns)` is in fact 2-arg, and `ReadNamespaceConfig` + `ExtractMode/ExtractMTLSMode/ExtractTLSBridgeMode/ExtractEgressEnforcement` all exist as documented.
- **Generated artifacts are correct.** deepcopy and CRD regenerated; `charts/kagenti-operator/crds/...` is byte-identical to `config/crd/bases/...` (Helm sync correct).
- **e2e test change is an improvement.** The old `It("should produce a different config-hash")` description contradicted its own assertion (`Expect(overridesHash).To(Equal(minimalHash))`). This PR fixes the description to match the assertion and swaps the now-removed `identity` fixture for a live `mtlsMode: strict` field. The passing E2E run confirms `mtlsMode` is excluded from the config hash.

**Areas reviewed:** Go API types, generated deepcopy/CRD, Helm CRD sync, docs (api-reference, authbridge-webhook, controller-webhook-interaction), controller + e2e tests, sample CRs.
**Commits:** 1, signed-off (DCO green).
**CI status:** all checks passing.

### Action needed before merge: rebase
The PR is currently `CONFLICTING` against `main`. There is exactly **one** real conflict — `kagenti-operator/docs/controller-webhook-interaction.md`; every other file auto-merges cleanly, so this is a low-risk rebase.

_Assisted-By: Claude Code_

Issue #410, opened 2026-06-06
Eliminate admin credentials from operator client registration


Alan-Cha:
## Problem

Currently, the operator uses Keycloak admin credentials to register OAuth clients for agents. This has several security issues:
- Admin credentials provide full realm permissions (create/delete users, realms, clients, roles)
- Long-lived credentials requiring manual rotation
- Operator compromise = full realm admin access
- Admin credentials stored as secrets in every agent namespace

## Solution

Use the operator's SPIFFE JWT-SVID to authenticate with Keycloak instead of admin credentials.

## Implementation

This epic is implemented across two PRs:

### Core Implementation
- **#349**: Operator-side SPIFFE authentication
  - JWT-SVID authentication with Keycloak Admin API
  - Feature flags and Helm configuration
  - Backward compatible (opt-in via `keycloak.spiffeAuth.enabled`)

### Platform Automation
- **kagenti/kagenti#1837**: Bootstrap automation and installation scripts
  - Automated Keycloak setup (SPIFFE IdP, operator client, role assignment)
  - Helm hook job for zero-manual-step deployment
  - Installation script integration (`ENABLE_OPERATOR_SPIFFE_AUTH=true`)

## Benefits

- ✅ Scoped permissions (manage-clients only, not full admin)
- ✅ No credential storage (cryptographic JWT-SVID authentication)
- ✅ Short-lived tokens (1 hour, auto-rotated)
- ✅ Audit trail shows operator SPIFFE ID

## Related

- Builds on: #320 (operator reads admin creds from kagenti-system)
- Related: kagenti/kagenti#1422 (Helm chart changes for #320)

Assisted-By: Claude Code

helen-stanton:
@Alan-Cha checking in if this issue is still in progress - and if so which milestone it is targetted to please. or should it be closed - thanks

Alan-Cha:
This has been completed

PR #349, opened 2026-05-11
Feat: SPIFFE authentication for operator client registration (Admin API)

Alan-Cha:
## Summary

Implements #1421 - Eliminate admin credentials from client registration using **SPIFFE JWT-SVID authentication with Keycloak Admin API**.

This PR provides the operator-side implementation. Platform bootstrap automation is in kagenti/kagenti#2135.

## Problem

Currently, the operator uses admin credentials to register OAuth clients via Keycloak Admin API. This has several security issues:
- Admin credentials = full realm permissions (create/delete users, realms, clients, roles)
- Long-lived credentials requiring manual rotation
- Operator compromise = full realm admin access
- Admin credentials stored as secrets in every agent namespace

## Solution

Use the operator's **SPIFFE JWT-SVID** to authenticate with Keycloak Admin API using **federated-jwt client authenticator**.

### Architecture

```
Operator Pod
├─> spiffe-helper sidecar (standalone container in operator pod)
│   └─> writes JWT-SVID to /opt/jwt_svid.token (auto-rotates)
├─> Manager reads JWT-SVID from file
└─> Authenticates to Keycloak:
    POST /realms/kagenti/protocol/openid-connect/token
    - grant_type: client_credentials
    - client_assertion_type: urn:ietf:params:oauth:client-assertion-type:jwt-spiffe
    - client_assertion: <JWT-SVID>
    → Returns access token with manage-clients role
    → Use same Admin API endpoints as before
```

### Benefits

| Current (Admin Creds) | With SPIFFE Auth (This PR) |
|----------------------|----------------------|
| Admin username/password | JWT-SVID (cryptographic) |
| Full realm admin permissions | Scoped (manage-clients only) |
| Long-lived credentials | Short-lived (1 hour), auto-rotates |
| Manual rotation | Automatic (spiffe-helper) |
| Secrets in all namespaces | No secrets |
| Audit trail: "admin" user | Audit trail: operator SPIFFE ID |

### Core Changes

**internal/keycloak/admin.go**:
- Added `JWTSVIDGrantToken()` method for JWT-SVID authentication
- Uses `client_assertion_type: urn:ietf:params:oauth:client-assertion-type:jwt-spiffe`
- Returns access token with operator's assigned roles

**internal/controller/clientregistration_controller.go**:
- Added SPIFFE auth fields: `UseSpiffeAuth`, `OperatorClientID`, `JWTSVIDPath`, `Recorder`
- Dual authentication path:
[...]

PR #473, opened 2026-07-02
Feat(operator): implement SPIFFE JWT-SVID authentication for client registration

Alan-Cha:
## Summary

Implements #1421 — eliminates admin credentials from operator client registration
by using the operator's SPIFFE identity (JWT-SVID) to authenticate with the
Keycloak Admin API.

## Problem

The operator currently uses admin credentials to register OAuth clients. This means:
- Admin credentials stored as secrets, requiring manual rotation
- Operator compromise = full realm admin access
- Long-lived credentials in the cluster

## Solution

The operator authenticates with its SPIFFE JWT-SVID via Keycloak's `federated-jwt`
client authenticator, using only the scoped `manage-clients` role.

## Architecture

```
Operator Pod
├─> spiffe-helper sidecar (standalone container)
│   └─> writes JWT-SVID to /opt/jwt_svid.token (auto-rotates)
├─> Manager reads JWT-SVID from file
└─> POST /realms/kagenti/protocol/openid-connect/token
    client_assertion_type: urn:ietf:params:oauth:client-assertion-type:jwt-spiffe
    → access token with manage-clients role
    → same Admin API endpoints as before
```

### Go (`internal/keycloak/admin.go`)
New `JWTSVIDGrantToken()` method authenticates the operator using JWT-SVID.

### Go (`internal/controller/clientregistration_controller.go`)
Dual authentication path:
- `UseSpiffeAuth=true`: reads JWT-SVID from file, calls `JWTSVIDGrantToken()`
- `UseSpiffeAuth=false`: admin credentials (legacy, default)

Security: path traversal protection (`filepath.Clean` + whitelist), JWT-SVID
never logged, all failures surface as Kubernetes Events.

### Go (`cmd/main.go`)
Registers `--use-spiffe-auth`, `--jwt-svid-path`, `--operator-client-id` flags.

### Helm (`charts/kagenti-operator/`)
- `configmap-spiffe-helper.yaml` (new): configures spiffe-helper to write
  JWT-SVID to `/opt/jwt_svid.token`. Set `spiffe.operatorAuth.jwtAudience`
  explicitly — the fallback uses `keycloak.publicUrl` which defaults to empty.
- `manager.yaml`: conditionally adds spiffe-helper sidecar (UID 65532, matching
  the manager so the 600-mode file is readable), jwt-svid emptyDir volume, SPIFFE
  flags — all gated on `spiffe.operatorAuth.enabled`.
- `values.yaml`: new `spiffe:` block, both `enabled` and `operatorAuth.enabled`
  default to `false`.

## Security improvements

| Before | After |
|---|---|
| Admin credentials (full realm) | JWT-SVID (cryptographic, short-lived) |
| `manage-*` roles | `manage-clients` only |
| Manual rotation | Auto-rotates (spiffe-helper) |
| Secrets in namespace | No secrets |

## E2E testing

Complete — multiple fresh-cluster runs, all 9 steps passed.
See https://gist.github.com/Alan-Cha/527be60c79b40aedd54fe7be586c3452

## Related

- Implements #410
- Platform bootstrap: kagenti/kagenti#2141 (companion clean PR)
- Replaces: #349 (superseded by this clean PR)

## Checklist

- [x] All commits signed-off (DCO)
- [x] Feature gated (default: disabled)
- [x] Backward compatible — `UseSpiffeAuth=false` uses admin credentials

Assisted-By: Claude Code (Anthropic AI) <noreply@anthropic.com>

### DIFF
--- a/charts/kagenti-operator/templates/manager/configmap-spiffe-helper.yaml
+++ b/charts/kagenti-operator/templates/manager/configmap-spiffe-helper.yaml
@@ -18,7 +18,7 @@ data:
     svid_file_name = ""
     svid_bundle_file_name = ""
     jwt_svids = [{
-        jwt_audience = "{{ .Values.spiffe.operatorAuth.jwtAudience | default (printf "%s/realms/%s" .Values.keycloak.publicUrl .Values.keycloak.realm) }}"
+        jwt_audience = "{{ required "spiffe.operatorAuth.jwtAudience must be set when spiffe.operatorAuth.enabled=true (e.g. http://keycloak.example.com/realms/kagenti)" .Values.spiffe.operatorAuth.jwtAudience }}"
         jwt_svid_file_name = "/opt/jwt_svid.token"
     }]
 {{- end }}
--- a/charts/kagenti-operator/templates/manager/configmap-spiffe-helper.yaml
+++ b/charts/kagenti-operator/templates/manager/configmap-spiffe-helper.yaml
@@ -1,0 +1,24 @@
+{{- if and .Values.spiffe .Values.spiffe.enabled .Values.spiffe.operatorAuth .Values.spiffe.operatorAuth.enabled }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: operator-spiffe-helper-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "chart.labels" . | nindent 4 }}
+data:
+  config.hcl: |
+    agent_address = "/spiffe-workload-api/spire-agent.sock"
+    # cmd/cmd_args fire on X.509 SVID renewals only, not JWT SVIDs.
+    # File permissions are handled by running spiffe-helper as the same UID
+    # as the manager container (65532) so it can read the 600-mode file.
+    renew_signal = ""
+    cert_dir = ""
+    svid_file_name = ""
+    svid_bundle_file_name = ""
+    jwt_svids = [{
+        jwt_audience = "{{ .Values.spiffe.operatorAuth.jwtAudience | default (printf "%s/realms/%s" .Values.keycloak.publicUrl .Values.keycloak.realm) }}"
+        jwt_svid_file_name = "/opt/jwt_svid.token"
+    }]
+{{- end }}
--- a/charts/kagenti-operator/templates/manager/manager.yaml
+++ b/charts/kagenti-operator/templates/manager/manager.yaml
@@ -92,7 +92,7 @@ spec:
             - "--credential-wait-timeout={{ .Values.authbridgeConfig.credentialWaitTimeout }}"
             {{- end }}
             {{- end }}
-{{- if .Values.sigstore.cardVerification.enabled }}
+            {{- if .Values.sigstore.cardVerification.enabled }}
             - "--enable-sigstore-verification=true"
             {{- if .Values.sigstore.cardVerification.auditMode }}
             - "--sigstore-audit-mode=true"
--- a/charts/kagenti-operator/templates/manager/manager.yaml
+++ b/charts/kagenti-operator/templates/manager/manager.yaml
@@ -193,8 +193,8 @@ spec:
             {{- end }}
         {{- if and .Values.spiffe .Values.spiffe.enabled .Values.spiffe.operatorAuth .Values.spiffe.operatorAuth.enabled }}
         - name: spiffe-helper
-          image: ghcr.io/kagenti/kagenti-extensions/spiffe-helper:latest
-          imagePullPolicy: IfNotPresent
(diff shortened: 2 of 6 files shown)


charts/kagenti-operator/templates/manager/manager.yaml
```diff
@@ -181,6 +186,42 @@ spec:
...
               readOnly: true
             {{- end }}
+            {{- if and .Values.spiffe .Values.spiffe.enabled .Values.spiffe.operatorAuth .Values.spiffe.operatorAuth.enabled }}
+            - name: jwt-svid
+              mountPath: /opt
+              readOnly: true
+            {{- end }}
+        {{- if and .Values.spiffe .Values.spiffe.enabled .Values.spiffe.operatorAuth .Values.spiffe.operatorAuth.enabled }}
+        - name: spiffe-helper
+          image: ghcr.io/kagenti/kagenti-extensions/spiffe-helper:latest
```
esnible: **suggestion:** The spiffe-helper sidecar is pinned to `:latest`. Every other image in this chart is templated/pinned (see the manager container just above, which uses `.Values.controllerManager.container.image.repository`/`.tag`). `:latest` is non-reproducible — pull behavior varies by node cache and can silently change the sidecar under a running pod, which is also a supply-chain concern for a component in the auth path. Recommend adding a templated `repository`/`tag` (ideally defaulting to a digest or version tag) under the `spiffe.operatorAuth` values block.

charts/kagenti-operator/templates/manager/configmap-spiffe-helper.yaml
```diff
@@ -0,0 +1,24 @@
...
+    agent_address = "/spiffe-workload-api/spire-agent.sock"
+    # cmd/cmd_args fire on X.509 SVID renewals only, not JWT SVIDs.
+    # File permissions are handled by running spiffe-helper as the same UID
+    # as the manager container (65532) so it can read the 600-mode file.
+    renew_signal = ""
+    cert_dir = ""
+    svid_file_name = ""
+    svid_bundle_file_name = ""
+    jwt_svids = [{
+        jwt_audience = "{{ .Values.spiffe.operatorAuth.jwtAudience | default (printf "%s/realms/%s" .Values.keycloak.publicUrl .Values.keycloak.realm) }}"
```
esnible: **suggestion:** The `jwt_audience` fallback is `printf "%s/realms/%s" .Values.keycloak.publicUrl .Values.keycloak.realm`, but `keycloak.publicUrl` defaults to `""`. So enabling SPIFFE auth without setting `jwtAudience` yields an audience of `/realms/kagenti`, which fails token validation with a confusing runtime error rather than at install time. The PR body already calls this out. Consider `{{ required "spiffe.operatorAuth.jwtAudience or keycloak.publicUrl must be set" ... }}` so the misconfiguration surfaces at `helm install` instead of a failing reconcile loop.

esnible (approved):
Clean, well-scoped implementation of JWT-SVID operator auth. The dual-auth split is correct: both paths converge on `RegisterOrFetchClientWithToken` with identical params, the SPIFFE path is fully feature-gated (`spiffe.enabled` + `spiffe.operatorAuth.enabled`, both default `false`), and legacy admin-credential auth is untouched as the default.

Verified against the branch:
- All reused helpers (`trimBaseURL`, `truncate`, `httpc`, `adminTokenResponse`) and imports already exist in `admin.go`.
- `corev1` is imported and `owner` is in scope for all `Recorder.Event` calls.
- The `spiffe-workload-api` CSI volume guard is correctly widened to `or verifiedFetch.enabled (spiffe operatorAuth)`, so the sidecar's mount is satisfied.
- Path-traversal protection (`filepath.Clean` + `/opt/`, `/var/run/secrets/` prefix whitelist) is sound against `..` sequences.
- Error classification is appropriate: permanent failure on bad config path, transient requeue on read/auth failure. JWT-SVID is never logged.

All 16 CI checks green (including E2E), all commits signed-off. Two non-blocking suggestions inline.

Assisted-By: Claude Code

charts/kagenti-operator/templates/manager/manager.yaml
Alan-Cha: Fixed in commit `541a618` — added `spiffe.operatorAuth.spiffeHelper.image.{repository,tag,pullPolicy}` values defaulting to `ghcr.io/kagenti/kagenti-extensions/spiffe-helper:v0.6.0-alpha.4` (latest kagenti-extensions release). The template now uses those values instead of the hardcoded `:latest`.

charts/kagenti-operator/templates/manager/configmap-spiffe-helper.yaml
Alan-Cha: Fixed in commit `541a618` — replaced the broken `printf` fallback with `required`, so `helm install` fails immediately with a clear message if `jwtAudience` is unset rather than silently producing `/realms/kagenti` at runtime.

PR #483, opened 2026-07-15
Fix(controller,webhook): write client-id.txt in federated-jwt mode for inbound JWT validation

### LINKED ISSUE #482: bug: inbound JWT validation broken in federated-jwt mode (client-id.txt never written)
## Problem

When `CLIENT_AUTH_TYPE=federated-jwt`, agent pods reject all authenticated inbound requests with HTTP 503:

```
identity not yet configured (credentials pending)
```

The `authbridge-proxy` sidecar polls indefinitely for `/shared/client-id.txt` and logs:

```
credential file still not available path=/shared/client-id.txt waited=4h12m0s
```

## Root cause

PR #476 correctly skips writing `client-secret.txt` in federated-jwt mode — agents authenticate outbound via JWT-SVID, so no Keycloak client secret is needed. However, the change skipped the entire `ensureClientCredentialsSecret` call, which also writes `client-id.txt`. It also skipped `patchTemplate`, so the credential Secret is never mounted into the pod at all.

`client-id.txt` is read by the authbridge inbound `jwt-validation` plugin to determine the expected audience when validating JWTs on incoming requests. Without it, the plugin stays in its init loop forever and rejects all inbound traffic with 503.

The broken chain in federated-jwt mode:
1. `ensureClientCredentialsSecret` skipped → no Secret, no `client-id.txt`
2. `patchTemplate` skipped → no Secret annotation on pod template
3. Webhook skips pre-populating the annotation → Secret never mounted

## Fix

In `ensureClientCredentialsSecret`, make `client-secret.txt` conditional:

```go
if clientSecret != "" {
    sec.StringData["client-secret.txt"] = clientSecret
}
sec.StringData["client-id.txt"] = clientID
```

In the controller reconcile loop, remove the `authType != "federated-jwt"` guard so both `ensureClientCredentialsSecret` (called with empty `clientSecret`) and `patchTemplate` still run in federated-jwt mode. The resulting Secret contains only `client-id.txt` (the agent's SPIFFE ID), which is all the inbound validator needs.

## Impact

All agent deployments with `CLIENT_AUTH_TYPE=federated-jwt` — inbound requests return 503 until this is fixed.

Alan-Cha:
## Problem

In `federated-jwt` mode, agent pods reject all inbound authenticated requests with HTTP 503:

```
identity not yet configured (credentials pending)
```

The `authbridge-proxy` sidecar polls indefinitely for `/shared/client-id.txt` — the file the inbound `jwt-validation` plugin reads to determine the expected audience when validating incoming JWTs.

## Root cause

PR #476 correctly intended to skip writing `client-secret.txt` in federated-jwt mode. However it skipped the entire `ensureClientCredentialsSecret` call, which also skipped `client-id.txt`. It also skipped `patchTemplate`, so the Secret annotation was never set and the webhook never mounted the Secret into the pod at all. The broken chain:

1. `ensureClientCredentialsSecret` skipped → no Secret, no `client-id.txt`
2. `patchTemplate` skipped → no annotation on pod template → webhook doesn't mount Secret
3. Webhook separately skipped annotation pre-population → same result

## Fix

**Controller** — in federated-jwt mode, pass an empty `clientSecret` to `ensureClientCredentialsSecret` instead of skipping the call. The Secret is always written with both keys: `client-id.txt` (the agent's SPIFFE ID) and `client-secret.txt` (empty string in federated-jwt mode). Writing an empty `client-secret.txt` rather than omitting the key avoids a Kubernetes quirk where a subPath mount for an absent Secret key creates a directory at the mount path instead of a file. Remove the `authType` guard so `patchTemplate` also runs in federated-jwt mode.

**Webhook** — remove the federated-jwt guard on annotation pre-population. The Secret now exists in this mode, so the volume mount annotation is needed for the first pod to start without a pod recreate.

In federated-jwt mode AuthBridge uses `identity.type=spiffe` for outbound token exchange and never reads `client-secret.txt`, so the empty value is harmless.

## Testing

- `go build ./...` clean
- `go vet ./internal/controller/... ./internal/webhook/...` clean

Fixes #482

Assisted-By: Claude Code

### DIFF
--- a/kagenti-operator/internal/webhook/v1alpha1/authbridge_webhook.go
+++ b/kagenti-operator/internal/webhook/v1alpha1/authbridge_webhook.go
@@ -97,22 +97,20 @@ func (w *AuthBridgeWebhook) Handle(ctx context.Context, req admission.Request) a
 	// Without this, the first pod comes up with an empty /shared/ and envoy returns 503
 	// "identity not yet configured (credentials pending)" until the user deletes the pod.
 	//
-	// Skip in federated-jwt mode: AuthBridge reads JWT-SVIDs from the SPIFFE workload API
-	// socket directly and no credential Secret is created or needed. Injecting the volume
-	// mount for a non-existent Secret would leave pods stuck in Init.
+	// In federated-jwt mode the Secret still exists (written by the operator with only
+	// client-id.txt, no client-secret.txt) so the volume mount is still needed. The
+	// inbound jwt-validation plugin reads client-id.txt to determine the expected audience;
+	// without the mount the plugin polls forever and rejects all inbound traffic with 503.
 	if pod.Annotations[injector.AnnotationKeycloakClientSecretName] == "" &&
 		clientreg.WorkloadWantsOperatorClientReg(pod.Labels, w.Mutator.GetFeatureGates().InjectTools) {
-		nsConfig, _ := injector.ReadNamespaceConfig(ctx, w.Mutator.APIReader, req.Namespace)
-		if nsConfig == nil || nsConfig.ClientAuthType != injector.ClientAuthTypeFederatedJWT {
-			if pod.Annotations == nil {
-				pod.Annotations = map[string]string{}
-			}
-			pod.Annotations[injector.AnnotationKeycloakClientSecretName] =
-				clientreg.KeycloakClientCredentialsSecretName(req.Namespace, resourceName)
-			authbridgelog.Info("pre-populated Keycloak client credentials annotation",
-				"namespace", req.Namespace, "name", resourceName,
-				"secret", pod.Annotations[injector.AnnotationKeycloakClientSecretName])
+		if pod.Annotations == nil {
+			pod.Annotations = map[string]string{}
 		}
+		pod.Annotations[injector.AnnotationKeycloakClientSecretName] =
+			clientreg.KeycloakClientCredentialsSecretName(req.Namespace, resourceName)
+		authbridgelog.Info("pre-populated Keycloak client credentials annotation",
+			"namespace", req.Namespace, "name", resourceName,
+			"secret", pod.Annotations[injector.AnnotationKeycloakClientSecretName])
 	}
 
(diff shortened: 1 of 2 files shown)


kagenti-operator/internal/webhook/v1alpha1/authbridge_webhook.go
```diff
@@ -97,22 +97,20 @@ func (w *AuthBridgeWebhook) Handle(ctx context.Context, req admission.Request) a
 	// Without this, the first pod comes up with an empty /shared/ and envoy returns 503
 	// "identity not yet configured (credentials pending)" until the user deletes the pod.
 	//
-	// Skip in federated-jwt mode: AuthBridge reads JWT-SVIDs from the SPIFFE workload API
-	// socket directly and no credential Secret is created or needed. Injecting the volume
-	// mount for a non-existent Secret would leave pods stuck in Init.
+	// In federated-jwt mode the Secret still exists (written by the operator with only
```
pdettori, line 100: nit: this comment says the Secret is "written by the operator with only client-id.txt, no client-secret.txt", but the controller change in this same PR writes `client-secret.txt` as an empty string rather than omitting it (that's the whole point of the subPath fix). Suggest rewording to "written with client-id.txt and an empty client-secret.txt" so it matches the actual behavior and the controller-side comment.

pdettori (approved):
Correct, minimal fix that restores the inbound-JWT audience config broken by #476.

Verified the logic against the full head-branch files: the controller now always calls `ensureClientCredentialsSecret` + `patchTemplate` (passing an empty `clientSecret` in federated-jwt mode), and the webhook now always pre-populates the annotation. The two sides are consistent — the Secret is created for every auth type and the webhook annotates for every eligible workload. The subPath-creates-a-directory-for-absent-key rationale for writing an empty `client-secret.txt` is a real Kubernetes quirk and a sound defensive choice.

Supply-chain gate clear (no `.claude/`/`.vscode/` changes). All 16 CI checks green, both commits signed off.

One non-blocking nit inline.

Assisted-By: Claude Code

PR #501, opened 2026-08-03
Fix: 🐛 Use upstream ghcr.io/spiffe/spiffe-helper image for operator sidecar

### LINKED ISSUE #2334: Operator SPIFFE auth: ghcr.io/rossoctl/cortex/spiffe-helper:latest sidecar image unpublished/inaccessible (403)
## Summary

Even after working around #2333 (unpublished `operator-spiffe-bootstrap` image), the operator's own pod cannot run with `operator-chart.spiffe.operatorAuth.enabled=true` because its `spiffe-helper` sidecar container fails to pull:

```
ghcr.io/rossoctl/cortex/spiffe-helper:latest
```

## Root cause

Same failure signature as #2333 — anonymous pull returns `403 Forbidden`:

```
Failed to pull image "ghcr.io/rossoctl/cortex/spiffe-helper:latest":
failed to resolve reference "ghcr.io/rossoctl/cortex/spiffe-helper:latest":
failed to authorize: failed to fetch anonymous token: ...403 Forbidden
```

This sidecar is what writes the operator's own JWT-SVID to `/opt/jwt_svid.token` (per `docs/users-guides/authentication.md`'s architecture diagram) — without it, the operator has no way to actually authenticate to Keycloak with SPIFFE at runtime, even if the bootstrap Job (which only registers the Keycloak client) succeeds.

## Impact

`rossoctl-controller-manager` pod stays `1/2 Ready` indefinitely (`ImagePullBackOff` on the `spiffe-helper` container). The `manager` container itself comes up fine, but operator SPIFFE auth is non-functional at runtime.

## Verified fix works with a substitute image

The container's args (`-config /etc/spiffe-helper/config.hcl`) match the upstream open-source `spiffe/spiffe-helper` CLI exactly, suggesting `ghcr.io/rossoctl/cortex/spiffe-helper` is just a rebrand/republish of that tool that was never actually pushed. I retagged the real upstream image (`ghcr.io/spiffe/spiffe-helper:nightly`) as `ghcr.io/rossoctl/cortex/spiffe-helper:latest`, loaded it into the Kind cluster (container's `imagePullPolicy` is already `IfNotPresent`, no patch needed), and the operator came up 2/2 Ready with the sidecar logging `"JWT SVID updated"`. The operator then logged `"SPIFFE ID authentication enabled: using JWT-SVID for client registration"` and successfully registered test workloads (weather-service/weather-tool) as Keycloak clients using its SPIFFE identity, with no `keycloak-admin-secret` access.

## Suggested fix

Publish `ghcr.io/rossoctl/cortex/spiffe-helper` (likely from the `cortex` repo's build pipeline) so operator SPIFFE auth is usable out of the box, or point the operator-chart default at the upstream `ghcr.io/spiffe/spiffe-helper` image directly if no rossoctl-specific fork is actually needed.

## Related

- #2333 (operator-spiffe-bootstrap image, same root cause pattern)

## Environment

- Tag: `v0.7.0-alpha.6`
- `operator-chart` version: `0.3.0-alpha.10`
- Local Kind cluster, macOS/Colima

Alan-Cha:
## Summary

The operator's own `spiffe-helper` sidecar (injected into `rossoctl-controller-manager` when `spiffe.operatorAuth.enabled=true`) points at `ghcr.io/rossoctl/cortex/spiffe-helper:latest`, which was never published (`403 Forbidden` on pull). The operator pod gets stuck at `1/2 Ready` and never obtains a JWT-SVID, so operator SPIFFE auth never actually authenticates at runtime — even though the bootstrap Job (separately, in rossoctl/rossoctl) successfully registers the Keycloak client.

The container's CLI args (`-config /etc/spiffe-helper/config.hcl`) match the real upstream SPIFFE project tool exactly, and an earlier branch already used `ghcr.io/spiffe/spiffe-helper:0.11.0` before this regressed to the broken `rossoctl/cortex` path.

Key changes:
- `charts/operator/templates/manager/manager.yaml:196` — swap `ghcr.io/rossoctl/cortex/spiffe-helper:latest` → `ghcr.io/spiffe/spiffe-helper:0.11.0`.

Pinned to a specific version rather than `:latest`: unlike `authbridge`/`envoyProxy`/`authbridgeLite`/`proxyInit` (rossoctl-built images that default to `:latest` at this chart level and get pinned downstream by the `rossoctl/rossoctl` consumer chart), `spiffe-helper` is a genuine third-party upstream dependency with no downstream pin step — so it's tagged directly here to the current latest stable release (`0.11.0`, confirmed via `docker manifest inspect`).

Verified via `helm template` that the rendered `spiffe-helper` container now uses the correct pinned image, and confirmed on a live Kind cluster (loading a substitute image manually) that the operator comes up `2/2 Ready` and logs `"SPIFFE ID authentication enabled: using JWT-SVID for client registration"` once this image is resolvable.

## Related issue(s)

Fixes rossoctl/rossoctl#2334

### DIFF
--- a/charts/operator/templates/manager/manager.yaml
+++ b/charts/operator/templates/manager/manager.yaml
@@ -193,7 +193,7 @@ spec:
             {{- end }}
         {{- if and .Values.spiffe .Values.spiffe.enabled .Values.spiffe.operatorAuth .Values.spiffe.operatorAuth.enabled }}
         - name: spiffe-helper
-          image: ghcr.io/spiffe/spiffe-helper:latest
+          image: ghcr.io/spiffe/spiffe-helper:0.11.0
           imagePullPolicy: IfNotPresent
           args:
             - "-config"
--- a/charts/operator/templates/manager/manager.yaml
+++ b/charts/operator/templates/manager/manager.yaml
@@ -193,7 +193,7 @@ spec:
             {{- end }}
         {{- if and .Values.spiffe .Values.spiffe.enabled .Values.spiffe.operatorAuth .Values.spiffe.operatorAuth.enabled }}
         - name: spiffe-helper
-          image: ghcr.io/rossoctl/cortex/spiffe-helper:latest
+          image: ghcr.io/spiffe/spiffe-helper:latest
           imagePullPolicy: IfNotPresent
           args:
             - "-config"


evaline-ju (commented):
q - should we avoid "latest" tags for images, especially if they're upstream/out of our control?

Alan-Cha:
@evaline-ju Yes, you are correct, but I noticed that most of the images in the operator are using the latest tag like https://github.com/rossoctl/operator/blob/47e7e301fc5c7494a6098b68bc6d51def4786ddf/charts/operator/values.yaml#L249

However, as you mentioned, this is an external dependency so I think you're right to point out that we should tag this one.

Alan-Cha:
The plan is to remove this dependency soon, but as of now, it's pointing at the wrong image

evaline-ju (approved):
lgtm!

PR #506, opened 2026-08-11
Fix: stop rendering duplicate rossoctl-authbridge SCC (rossoctl#2354)

cwiklik:
## Summary

Fixes the OpenShift install-blocker **rossoctl/rossoctl#2354**.

This chart rendered a second `rossoctl-authbridge` SCC (+ its `use` ClusterRole) that **duplicates and conflicts** with the ones the rossoctl **platform chart** has shipped since 2026-03-27. In a combined rossoctl install (this chart is pulled as a subchart), two docs declare the same cluster-scoped SCC → Helm merge-patch adoption fails on OpenShift → **install hard-blocks**. Even when de-duped to *this* chart's spec, agent pods then fail admission because this spec is restrictive (`MustRunAsNonRoot`, `allowedCapabilities:[]`, drop `ALL`) and can't admit the injected `proxy-init` (NET_ADMIN/NET_RAW, root).

The platform chart's SCC is a **strict superset** — identical `volumes` (incl. `csi` for SPIRE), and more permissive only where `proxy-init` requires it — so dropping this chart's copy loses nothing and also fixes the admission failure.

## Change

`charts/operator/templates/rbac/authbridge-scc.yaml`:
- **Remove** the `SecurityContextConstraints` and the `system:openshift:scc:rossoctl-authbridge` ClusterRole (both owned by the platform chart).
- **Keep** the `ClusterRoleBinding` granting the operator SA `use` of the SCC — its `roleRef` resolves to the platform-provided ClusterRole of the same name, so `ensureNamespaceSCCBinding`'s per-namespace delegation keeps working without an RBAC privilege-escalation error.

Net in a combined install: exactly one (permissive, platform) SCC + one ClusterRole; operator retains delegation. No new permissions vs v0.6.1.

## Standalone note (maintainer decision)

If this chart is meant to be installed **standalone on OpenShift** (without the platform chart), it still needs its own SCC. In that case, prefer gating the removed objects behind a `scc.create` value (default `true`) that rossoctl sets `false`, and make the standalone SCC spec permissive — rather than deleting outright. Happy to switch to that approach if standalone-via-Helm on OCP is a supported path. (The kustomize `config/security/rossoctl-authbridge-scc.yaml` path is unaffected.)

## Rollout

After merge: cut a new operator-chart release → re-pin `operator-chart` in `rossoctl` `charts/rossoctl/Chart.yaml` + `helm dependency update`.

Refs rossoctl/rossoctl#2354. Context (regression trace + field-by-field SCC comparison) in that issue.

Assisted-By: Claude Code

### DIFF
--- a/charts/operator/templates/rbac/authbridge-scc.yaml
+++ b/charts/operator/templates/rbac/authbridge-scc.yaml
@@ -1,62 +1,12 @@
 {{- if and .Values.rbac.enable (.Capabilities.APIVersions.Has "security.openshift.io/v1") }}
-# SCC for AuthBridge proxy-sidecar containers in agent namespaces.
-# Custom SCC needed because restricted-v2 does not permit CSI volumes
-# (csi.spiffe.io for SPIRE workload API).
-# groups is empty -- the operator creates per-namespace RoleBindings at runtime.
-apiVersion: security.openshift.io/v1
-kind: SecurityContextConstraints
-metadata:
-  name: rossoctl-authbridge
-  labels:
-    {{- include "chart.labels" . | nindent 4 }}
-    app.kubernetes.io/component: authbridge
-groups: []
-allowHostDirVolumePlugin: false
-allowHostIPC: false
-allowHostNetwork: false
-allowHostPID: false
-allowHostPorts: false
-allowPrivilegedContainer: false
-allowPrivilegeEscalation: false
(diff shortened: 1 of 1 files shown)


evaline-ju (approved):
might be good to highlight the standalone/maintainer decision point in the next release, especially if it affects current install paths?

PR #510, opened 2026-08-18
feat: Add spec.auth to AgentRuntime for SPIFFE token exchange

Alan-Cha:
## Summary

Implements issue #2174 by adding spec.auth field to the AgentRuntime CRD and operator logic to generate AuthBridge token-exchange routes from spec.auth.outbound configuration.

### CRD Extension
- Added spec.auth field with:
  - mode: Authentication mechanism (federated-jwt, client-secret, disabled)
  - outbound: List of routes defining destination matching and audiences for token exchange
- Generated CRD manifest with OpenAPI validation

### Operator Logic
- Modified webhook to fetch AgentRuntime CR for each workload
- Updated ensurePerAgentConfigMap to accept AgentRuntime parameter  
- Added route generation logic that reads spec.auth.outbound and injects routes into pipeline.outbound.plugins[token-exchange].config.routes
- Routes include destination matching (host or hostRegex) and SPIFFE audience list

### Tests
- Added pod_mutator_auth_test.go with 2 new tests:
  - TestEnsurePerAgentConfigMap_WithAuthRoutes: Verifies route injection
  - TestEnsurePerAgentConfigMap_NoAuth: Verifies backward compatibility
- Updated existing tests to pass nil for new agentRuntime parameter
- All webhook tests pass ✅

## Route Format

Generated routes are injected as:
```yaml
routes:
  - destination:
      host: "weather-tool-mcp.team1.svc.cluster.local"
    audiences:
      - "spiffe://localtest.me/ns/team1/sa/weather-tool"
```

## Backward Compatibility

- AgentRuntime without spec.auth continues to work (no routes injected)
- Nil AgentRuntime is handled gracefully (for workloads without CRs)

## Testing

Unit tests verify:
- Routes are correctly generated from spec.auth.outbound
- Both host and hostRegex matching work
- Multiple routes and multiple audiences per route work
- Backward compatibility preserved

Closes #2174

Assisted-By: Claude Code


operator/api/v1alpha1/agentruntime_types.go
```diff
@@ -155,6 +155,68 @@ type AgentRuntimeSpec struct {
...
+	// Typically includes the SPIFFE ID of the target service.
+	//
+	// Example: ["spiffe://localtest.me/ns/team1/sa/weather-tool"]
+	//
+	// +kubebuilder:validation:MinItems=1
+	Audiences []string `json:"audiences"`
+}
+
+// RouteMatch defines how to match an outbound destination.
+type RouteMatch struct {
```
cwiklik, line 198: suggestion: `RouteMatch` has both `host` and `hostRegex` as `+optional` with no "exactly one" constraint. An empty `RouteMatch` is accepted, and route-gen in pod_mutator.go then emits `destination: {}` into the token-exchange config — a silent catch-all/no-op in a credential-minting path (tokens exchanged for the listed audiences on *all* outbound calls, or none, depending on AuthBridge's empty-match semantics). Recommend a `+kubebuilder:validation:XValidation` (CEL) requiring exactly one of `host`/`hostRegex`, and defensively skip+WARN on an empty destination in route-gen. A test for the empty/both cases would lock it in.

operator/api/v1alpha1/agentruntime_types.go
```diff
@@ -155,6 +155,68 @@ type AgentRuntimeSpec struct {
...
+	// Example: "weather-tool-mcp.team1.svc.cluster.local"
+	//
+	// +optional
+	Host string `json:"host,omitempty"`
+
+	// HostRegex is a regex pattern to match hostnames.
+	// Example: ".*\\.team1\\.svc\\.cluster\\.local"
+	//
+	// +optional
+	HostRegex string `json:"hostRegex,omitempty"`
```
cwiklik, line 209: suggestion: `hostRegex` is an unbounded, arbitrary regex evaluated on the request path. Risks: ReDoS from a pathological pattern, and an over-broad pattern (`.*`) silently broadening which destinations get token-exchange for the configured audiences. Consider a `maxLength` bound and documenting that the pattern scopes credential exchange (broad regex = broad credential scope).

operator/internal/webhook/injector/pod_mutator.go
```diff
@@ -992,6 +1018,74 @@ func (m *PodMutator) ensurePerAgentConfigMap(
...
+								}
+								if outboundRoute.Destination.HostRegex != "" {
+									destination["hostRegex"] = outboundRoute.Destination.HostRegex
+								}
+								route["destination"] = destination
+
+								routes = append(routes, route)
+							}
+
+							pluginConfig["routes"] = routes
```
cwiklik, line 1078: nit: `pluginConfig["routes"] = routes` replaces any pre-existing `routes` key wholesale — if the namespace template ever ships default outbound routes, they'd be clobbered by the spec.auth-derived set. Assuming spec.auth is intended as the source of truth here, a one-line comment saying so would help future readers.

cwiklik (approved):
Adds `spec.auth` to AgentRuntime and generates AuthBridge token-exchange routes into the Envoy outbound pipeline (#2174). The route-gen is nicely defensive — nil-checked navigation of the untyped pipeline with WARN-and-skip at each level, gated to `mode: federated-jwt`, and a graceful missing-CR fallback (pre-CRD workloads get no routes, existing behavior preserved). `mode` is Enum-validated with a `client-secret` default, `audiences` is `MinItems=1`, and both generated CRD copies (charts/ and config/crd/bases/) are byte-identical.

Three hardening notes in the credential-minting path (all non-blocking — exploitability is gated by who can create AgentRuntime CRs, and the worst case is scope-broadening for the author's own audiences, not arbitrary escalation):
1. `RouteMatch` allows neither/both `host`+`hostRegex`; an empty match emits `destination: {}` into the token-exchange config (silent catch-all/no-op). Recommend a CEL XValidation requiring exactly one, plus a defensive skip+WARN.
2. `hostRegex` is unbounded — bound it (maxLength) given ReDoS + credential-scope-broadening.
3. `pluginConfig["routes"]` overwrite — confirm spec.auth is the intended source of truth.

15 checks green; E2E pending (maintainer trigger). All commits signed, no `.claude`/`.vscode`. LGTM.

Assisted-By: Claude Code

PR #511, opened 2026-08-18
feat: Add inboundInterception=transparent (drops port stealing)

huang195:
## Summary

Adds an opt-in inbound shape for `proxy-sidecar` / `lite` that **stops stealing the agent's port**. Companion to rossoctl/cortex#776 (**merged**), which implements the listener and iptables rules; this is the switch that turns them on. Until this lands, that code is inert — nothing sets `inbound_interception` or `INBOUND_TRANSPARENT_PORT`.

With `inboundInterception: transparent`, proxy-init installs a PREROUTING REDIRECT and AuthBridge recovers each connection's real destination via `SO_ORIGINAL_DST`, so the agent keeps the port it already binds.

## Why

Port stealing has three problems, all of which the existing code acknowledges:

1. **The relocated port is an unvalidated entrance.** `originalPort+1` is declared in the pod spec and directly reachable, so any pod can reach the agent without JWT validation. Pods are the granularity NetworkPolicy and ztunnel enforce at, so this was the boundary that mattered.
2. **It depends on the agent honoring `PORT`.** `pod_mutator.go` concedes that agents which hardcode their listen port "won't be affected" — in practice they collide with AuthBridge on the stolen port and the pod never starts. No config can fix that.
3. **Only `Ports[0]` of the first container with ports is relocated**, so a second declared port was never proxied. Undeclared ports aren't even in `usedPorts`, so `findFreePort` can hand the agent a port it already uses.

Transparent interception removes all three: no relocation, no `PORT` env var, no second port to discover, and every port the agent listens on is covered.

## Default is unchanged

Default stays `reverse-proxy`. Transparent costs a privileged proxy-init container, so it is chosen deliberately; with the field unset every existing path is byte-identical. Two fallbacks are deliberately biased toward the unprivileged shape:

- An unrecognized value falls back to `reverse-proxy`, so a typo cannot silently grant a `NET_ADMIN` init container.
- `transparent` + `egressEnforcement: none` falls back to `reverse-proxy`. The two features share one proxy-init container; without it nothing would REDIRECT to the inbound listener and inbound would be *silently* unenforced — worse than port stealing, which at least validates Service-routed traffic.

`proxy.allowedInboundInterception` lets a platform admin forbid it (`["reverse-proxy"]`) or mandate it (`["transparent"]`).

## Deliberately NOT adding an AgentRuntime spec field

The obvious move was to mirror `spec.egressEnforcement` — but **that field is dead surface**. The pod mutator has no access to the CR and resolves everything from the namespace ConfigMap, so nothing reads `spec.egressEnforcement` (`spec.mtlsMode` reaches the pod only as a rollout-triggering annotation, not as mutator input). An enum-validated spec field that silently does nothing is worse than no field, and plumbing CR → mutator properly is a separate change that would have to fix `mtlsMode` and `egressEnforcement` too.

**Worth deciding separately: `spec.egressEnforcement` should probably be either wired up or removed.**

## Details

- The per-agent ConfigMap gets `inbound_interception` + `transparent_inbound_addr` and must **not** get `reverse_proxy_addr`/`reverse_proxy_backend` — authbridge's config validation rejects that combination, so emitting both would crash-loop the pod.
- `SIDECAR_PORTS_EXCLUDE` carries the **resolved** forward-proxy port, not the script's `8081` default: `findFreePort` may have moved it, and an unexempted forward-proxy port would be swallowed by the inbound REDIRECT. Gating `9091` would put kubelet probes behind JWT validation.
- The sidecar declares its inbound port as `transparent-in`, so which shape is running is visible in the pod spec.
- Config validation rejects `transparentInboundPort == transparentPort`: both are listeners in one container, so a shared value fails the second bind at pod start, long after admission succeeded.

Upgrade-safe: the config loader overlays YAML onto compiled defaults, so an existing ConfigMap without the new keys keeps `8083` / both-allowed. Verified by running this operator against an unmodified 22-day-old cluster ConfigMap.

## Verification

- `make test` — 21 packages green, including 13 new injector tests and 3 new config-validation tests (table-driven, 9 further cases).
- No CRD changes (none needed); `make manifests generate` produces no drift.
- **Verified live on Kind**: agent keeps `:8000` with no `PORT` override, sidecar declares `transparent-in=8083` + `forward-proxy=8081`, proxy-init receives `INBOUND_TRANSPARENT_PORT=8083` / `POD_IP` (downward API) / `SIDECAR_PORTS_EXCLUDE=8081,9091,9093,9094`, and the per-agent ConfigMap contains `inbound_interception: transparent` + `transparent_inbound_addr: :8083` with no `reverse_proxy_*` keys.

## Port safety

`usedPorts` previously recorded the inbound port as reserved but never checked whether the agent had **declared** it. Under port stealing such a collision was accidentally survivable (`Ports[0]` got relocated); transparent mode deliberately removes that, so an agent declaring 8083 would have left two processes binding one port in a shared netns — admission succeeds, then the sidecar dies on `address already in use` in a container the user never wrote.

All sidecar-owned ports (8082, 8083, 9091, 9093, 9094) are now reserved unconditionally, and a declared collision on a port the sidecar *actually binds* falls back to port stealing with a warning naming the port and the owning listener. Reservation is separate from collision because `containerPort` is informational: under reverse-proxy nothing binds 8083, so declaring it is harmless.

This also closed a pre-existing hole — the transparent **egress** port (8082) was never reserved on main either, so an agent declaring 8081 could push the forward proxy onto a listener that is always on in this mode.

## Sequencing note

Default image tags are `:latest` (`values.yaml:250-252`, `defaults.go:38-45`). If a workload selects `transparent` while running an authbridge image predating cortex#776, the pod **crash-loops rather than silently bypassing**: the old binary ignores the unknown `inbound_interception` key, finds no `reverse_proxy_addr`/`reverse_proxy_backend` (deliberately omitted for transparent), and fails config validation. Loud rather than dangerous, and unreachable by accident since the default is `reverse-proxy` — but on a cluster pinned to an older tag, opting in needs the image bumped first.

## Verified end-to-end, with one gap

E2E suite **10 passed, 0 skipped** on Kind (rossoctl/rossoctl#2393). The operator-side injection was confirmed live: the agent keeps its port with no `PORT` override, the sidecar declares `transparent-in=8083`, proxy-init receives `INBOUND_TRANSPARENT_PORT` / `POD_IP` / `POD_IPS` / `SIDECAR_PORTS_EXCLUDE`, and the per-agent ConfigMap correctly omits `reverse_proxy_*`. The A/B that matters: the transparent agent's real port returns 401 while the reverse-proxy control's relocated port returns 200 — the bypass this closes.

**Still unverified: the ambient (HBONE) path's runtime behavior.** The cluster used runs only `istiod` — no ztunnel DaemonSet, no istio-cni — so there is no ambient data plane, and counters confirm all traffic took plain PREROUTING (`AB_INBOUND` REDIRECT 4 packets, ambient DNAT 0). The rules are verified installed and correctly ordered; whether ztunnel's re-originated connection matches `mark 0x539 + dst-type LOCAL` is still reasoned from redirect-mode precedent rather than observed. Needs a cluster with ztunnel deployed.


cwiklik (approved):
### Summary

Well-engineered, well-tested, **opt-in** feature that replaces port-stealing with transparent inbound interception (iptables PREROUTING REDIRECT + `SO_ORIGINAL_DST`), closing a real JWT-bypass hole — the relocated `originalPort+1` was a directly-reachable, unvalidated entrance. Default stays `reverse-proxy`, so no blast radius on merge. The PR body is exemplary: it enumerates the port-stealing failure modes it fixes and honestly flags what's still unverified.

**Verified correct:**
- **Dual-mode ConfigMap parity** — strict `if transparentInbound / else` in `listenerOverrides`: transparent emits `inbound_interception` + `transparent_inbound_addr` and never the `reverse_proxy_*` keys (which authbridge would reject together); reverse-proxy emits them. No path emits both or neither, and the ConfigMap, sidecar container, and proxy-init env all read one `transparentInbound` local, so the three artifacts can't disagree.
- **Port reservation** — all five sidecar ports (8082, 8083, 9091, 9093, 9094) reserved unconditionally in `usedPorts`; a declared agent collision on a port the sidecar actually binds falls back to port-stealing with a named warning.
- **Config validation** (`types.go:213-220`) — rejects `transparentInboundPort == transparentPort` and out-of-range.
- **Enum defaults fail-safe** — empty/unknown → `reverse-proxy`; `allowedInbound[0]` index is length-guarded; no unguarded map/slice access.
- **Feature-gated / off by default** — complies with the feature-flag policy.

**Non-blocking findings:**
1. **`SIDECAR_PORTS_EXCLUDE` omits 8082** (`pod_mutator.go:753` builds `forwardProxyPort,9091,9093,9094`). 8083 is correctly omitted (it's the REDIRECT target), but 8082 — the transparent *egress* listener you note is "always on in this mode" — is left out while 8081 (forward proxy) is excluded. Likely fine if the init script applies exclusions only to PREROUTING (8082 takes OUTPUT-originated traffic, never inbound), but the asymmetry with 8081 is worth confirming — it lands squarely in the ambient/HBONE path you flagged as unverified.
2. **Re-injection idempotency** — the transparent inbound env is added only when a proxy-init container is absent, so on re-admission the sidecar still binds the transparent listener and the ConfigMap still emits `transparent_inbound_addr`, but no PREROUTING rules get (re)installed → inbound silently unenforced. Mirrors the existing egress pattern, but the consequence (a bound listener nothing redirects to) is sharper here.
3. Your self-disclosed follow-ups — ambient/HBONE runtime still reasoned-not-observed (single-istiod cluster, ambient DNAT counter 0), `egressEnforcement` wire-or-remove, and the `:latest` sequencing (crash-loops loudly rather than bypassing on a pre-cortex#776 image) — are all legitimate follow-ups, not merge blockers, given default-off and the confirmed 401-on-real-port result.

Approving. The security posture strictly improves over port-stealing, the closed path is verified live, and the residual items are either opt-in-only or already flagged.

**Author**: huang195 (MEMBER — maintainer)
**Areas reviewed**: Go webhook injector/config, Helm values, dual-mode parity, port/iptables safety, tests
**CI**: 15 green; E2E Tests pending.

PR #512, opened 2026-08-19
fix(release): pin AuthBridge injection images + guard against :latest (#508)

### LINKED ISSUE #508: release.yml doesn't pin injected AuthBridge images (they ship as :latest)
## Problem
The operator chart pins the injected AuthBridge images (`authbridge`, `authbridge-envoy`, `authbridge-lite`, `proxy-init`) at **`:latest`** in `charts/operator/values.yaml`. The release workflow (`.github/workflows/release.yml`) only pins the **controller-manager** image (`yq … .controllerManager.container.image.tag = version`) — it never touches the injected images. So every published `operator-chart` (the whole 0.3 RC line: rc.1/rc.2/rc.3) shipped `:latest` for the AuthBridge images.

Non-reproducible, and against the no-`:latest`-in-release rule. It was invisible to both repos' pin checks (see rossoctl companion issue).

## What was done for v0.7.0 (symptom fix only)
`release-0.3` `values.yaml` was hand-pinned `:latest → ghcr.io/rossoctl/cortex/authbridge*:v0.6.0` (commit on release-0.3), so the shipped v0.3.0 chart and the v0.7.0 platform are clean. **This did not fix the process or `main`.**

## Fix (this issue)
1. Pin the injected images in `main`'s `values.yaml` too (so future releases don't regress), and/or
2. Have `release.yml` pin the injected images to a release-appropriate cortex tag, or **fail the build if any injected image is `:latest`**.

## Timing
Resolve **early in 0.8 — before the first 0.8 operator RC** — otherwise the 0.8 RCs regress to `:latest` and require hand-pinning again.

Assisted-By: Claude Code

cwiklik:
## Summary
The operator chart pinned the cortex-built AuthBridge injection images (`authbridge`, `authbridge-envoy`, `authbridge-lite`, `proxy-init`) at **`:latest`** in `charts/operator/values.yaml`. `release.yml` only pins the **controller-manager** image, so every published operator-chart shipped `:latest` for the injected images — non-reproducible (surfaced during the v0.7.0 GA cut).

## Changes
- **Pin** the 4 AuthBridge injection images `:latest` → **`v0.7.0-alpha.3`** (current cortex release; all 4 images published).
- **Guard** in `release.yml`: fail the release if any injected image is a floating tag (`:latest`/`:main`/`:master`) — so it can't regress silently.

Closes #508

Assisted-By: Claude Code

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -106,14 +106,33 @@ jobs:
     - name: Guard — injected AuthBridge images must be version-pinned (no floating tags)
       # release.yml pins the controller-manager image (below) but NOT the cortex-built
       # AuthBridge injection images; they were shipping at :latest (rossoctl/rossoctl#508).
-      # Fail the release if any is a floating tag so it can never regress silently.
+      # Positive assertion (tighter than a deny-list, which passes untagged/:dev/etc):
+      # every injection image must carry a :vX.Y.Z tag — in BOTH the chart values and
+      # the compiled Go fallbacks (config/defaults.go), which loader.go overlays the
+      # platform-config ConfigMap on top of, so a no-ConfigMap deploy (kustomize
+      # `make deploy`, webhook.enable=false) would otherwise still inject :latest.
+      # This is a release-time backstop; the same check runs per-PR in security-scans.yaml.
       run: |
-        cd ${{ env.CHARTS_PATH }}/operator
-        if grep -nE 'ghcr\.io/rossoctl/cortex/[a-z-]+:(latest|main|master)([^0-9a-zA-Z.-]|$)' values.yaml; then
-          echo "::error::injected AuthBridge image(s) pinned to a floating tag — pin to a cortex release tag before releasing (rossoctl/rossoctl#508)"
+        set -euo pipefail
+        fail=0
+        for k in envoyProxy authbridge authbridgeLite proxyInit; do
+          img=$(yq ".defaults.images.$k" ${{ env.CHARTS_PATH }}/operator/values.yaml)
+          if [[ ! "$img" =~ :v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
+            echo "::error::defaults.images.$k not version-pinned in values.yaml: $img"
+            fail=1
+          fi
+        done
+        while IFS= read -r img; do
+          if [[ ! "$img" =~ :v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
+            echo "::error::compiled default not version-pinned in config/defaults.go: $img"
+            fail=1
+          fi
+        done < <(grep -oE 'ghcr\.io/rossoctl/cortex/[a-z-]+:[^"]+' operator/internal/webhook/config/defaults.go)
+        if [ "$fail" -ne 0 ]; then
+          echo "::error::pin the flagged AuthBridge injection image(s) to a cortex release tag before releasing (rossoctl/rossoctl#508)"
           exit 1
         fi
-        echo "AuthBridge injected images are version-pinned OK"
+        echo "AuthBridge injection images are version-pinned OK (values.yaml + config/defaults.go)"
 
     - name: Package and push Helm chart
       run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -103,6 +103,18 @@ jobs:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
 
+    - name: Guard — injected AuthBridge images must be version-pinned (no floating tags)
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -103,6 +103,18 @@ jobs:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
 
+    - name: Guard — injected AuthBridge images must be version-pinned (no floating tags)
```
huang195, line 106: Guard sits in the `release` job, which is `needs: build-and-push` + `if: ref_type == 'tag'`. By the time it fails, `kagenti-operator` and `agentcard-signer` are already built and pushed to ghcr with the release tag — half-published release (images out, chart not), manual cleanup.

Two cheaper spots: a pre-flight job that `build-and-push` depends on, or the `Helm Chart Lint` job in `security-scans.yaml` which runs on every PR — then a `:latest` regression is caught at review time, not mid-release.

.github/workflows/release.yml
```diff
@@ -103,6 +103,18 @@ jobs:
...
+      # release.yml pins the controller-manager image (below) but NOT the cortex-built
+      # AuthBridge injection images; they were shipping at :latest (rossoctl/rossoctl#508).
+      # Fail the release if any is a floating tag so it can never regress silently.
+      run: |
+        cd ${{ env.CHARTS_PATH }}/operator
+        if grep -nE 'ghcr\.io/rossoctl/cortex/[a-z-]+:(latest|main|master)([^0-9a-zA-Z.-]|$)' values.yaml; then
```
huang195: Deny-list of 3 tag names + hardcoded registry path. These all pass the guard:

- no tag at all (`.../authbridge-envoy` → implicit `:latest`)
- `:dev` / `:edge` / `:nightly`
- any move off `ghcr.io/kagenti/kagenti-extensions/`

Positive assertion is strictly tighter, and `yq` is already installed in this job:

```bash
for k in envoyProxy authbridge authbridgeLite proxyInit; do
  img=$(yq ".defaults.images.$k" values.yaml)
  [[ "$img" =~ :v[0-9]+\.[0-9]+\.[0-9]+ ]] || { echo "::error::defaults.images.$k not version-pinned: $img"; exit 1; }
done
```

Keys the check on the 4 values that matter instead of on the registry string.

charts/operator/values.yaml
```diff
@@ -246,10 +246,10 @@ featureGates:
 # proxy-sidecar / lite mode (always-on enforce-redirect egress capture).
 defaults:
   images:
-    envoyProxy: ghcr.io/rossoctl/cortex/authbridge-envoy:latest
-    authbridge: ghcr.io/rossoctl/cortex/authbridge:latest
-    authbridgeLite: ghcr.io/rossoctl/cortex/authbridge-lite:latest
-    proxyInit: ghcr.io/rossoctl/cortex/proxy-init:latest
+    envoyProxy: ghcr.io/rossoctl/cortex/authbridge-envoy:v0.7.0-alpha.3
```
huang195, line 249: Chart layer pinned, compiled fallback not: `operator/internal/webhook/config/defaults.go:34-45` still has all four at `:latest`, and `loader.go:45` overlays the ConfigMap *on top of* `CompiledDefaults()` (fields absent from the file keep the compiled value). So any deploy without `kagenti-platform-config` — kustomize `make deploy`, or `webhook.enable=false` where the ConfigMap isn't rendered — still injects `:latest`. The new grep only reads `values.yaml`, so it will never catch it.

`defaults.go` already carries a `Keep in sync with charts/operator/values.yaml` comment for the spiffe-helper config; same treatment fits here.

Related: `operator/test/e2e/e2e_suite_test.go:66-69` still pulls `:latest` (and is missing `authbridge-lite`), so e2e never exercises the images the chart actually ships.

huang195 (approved):
Dobra robota — the pin is correct and verified: `v0.7.0-alpha.3` is cortex's current release (published 14 min before this PR) and cortex's `Build-Publish` matrix tags all four images with `github.ref_name` on tag push, run green. Guard logic is sound (`CHARTS_PATH` resolves, `grep` under `if` is errexit-safe).

Three non-blocking notes: the guard fires after the operator images are already pushed, it's a 3-tag deny-list rather than a positive pin assertion, and the compiled Go defaults in `defaults.go` still carry `:latest` where the grep can't reach them.

One question: the guard happily passes a prerelease tag, so operator v0.7.0 GA would ship cortex *alpha* images. Intended, or does the GA cut need a re-pin once cortex has a v0.7.0?

---

**Areas reviewed**: CI/GitHub Actions, Helm values · **Agent/IDE config (.claude/.vscode)**: none · **Commits**: 1, all signed-off: yes · **CI**: 16 pass, E2E pending

Assisted-By: Claude Code

PR #517, opened 2026-08-26
fix: generate AuthBridge routes in separate ConfigMap with correct format

### LINKED ISSUE #2334: Operator SPIFFE auth: ghcr.io/rossoctl/cortex/spiffe-helper:latest sidecar image unpublished/inaccessible (403)

Alan-Cha:
## Summary

Fixes two bugs in AuthBridge route generation from AgentRuntime `spec.auth.outbound`:

1. **Routes architecture bug**: Generate routes in separate ConfigMap with file reference instead of inline in config.yaml
2. **Routes format bug**: Generate routes in AuthBridge's `routing.Route` format instead of AgentRuntime CRD format

## Bug 1: Routes Architecture

**Problem**: AuthBridge crashed with "cannot unmarshal array into Go struct field" when routes were inline in config.yaml

**Root Cause**: AuthBridge's token-exchange plugin in proxy-sidecar mode expects routes loaded from an external file, not inline in the config

**Solution**:
- Generate routes in separate `authbridge-routes-<crName>` ConfigMap
- Reference via `routes: {file: "/etc/authproxy/routes.yaml"}` in config.yaml  
- Mount the routes ConfigMap as a volume

## Bug 2: Routes Format

**Problem**: Routes didn't match even after Bug 1 was fixed - AuthBridge logged "no matching route"

**Root Cause**: Operator generated routes in AgentRuntime CRD format, but AuthBridge expects `routing.Route` format

**Wrong format** (AgentRuntime CRD):
```yaml
- destination:
    host: "hostname"
  audiences:
    - "audience"
```

**Correct format** (AuthBridge routing.Route):
```yaml
- host: "hostname"
  target_audience: "audience"
```

**Solution**: Changed route generation to use flat AuthBridge format with:
- Flat `host` field (not nested under `destination`)
- Single `target_audience` string (not `audiences` array)

## Testing

Complete E2E test on fresh Kind cluster verified:
- ✅ Routes ConfigMap created with correct format
- ✅ Config references file correctly
- ✅ Routes match successfully (no "no matching route" errors)
- ✅ Token exchange plugin invoked correctly

## Files Modified

- `operator/internal/webhook/injector/pod_mutator.go`: Route generation and ConfigMap creation
- `operator/internal/webhook/injector/volume_builder.go`: Added `overrideRoutesConfigMapInVolumes` function

## Related Issues

Fixes rossoctl/rossoctl#2334

Assisted-By: Claude Code


operator/internal/webhook/injector/pod_mutator.go
```diff
@@ -1296,41 +1311,56 @@ func (m *PodMutator) ensurePerAgentConfigMap(
...
+			}
+			if outboundRoute.Destination.HostRegex != "" {
+				// AuthBridge router doesn't support hostRegex - use glob pattern in host field
+				route["host"] = outboundRoute.Destination.HostRegex
+			}
+
+			// target_audience is a single string, not array
+			// Take first audience if multiple specified
+			if len(outboundRoute.Audiences) > 0 {
+				route["target_audience"] = outboundRoute.Audiences[0]
```
clawgenti, line 1353: Silent data loss: if `outboundRoute.Audiences` has more than one entry, all but the first are silently dropped. Consider logging a warning here so operators know their intent wasn't fully applied:
```go
if len(outboundRoute.Audiences) > 1 {
    mutatorLog.Info("WARN: multiple audiences specified; only first used as target_audience",
        "namespace", namespace, "crName", crName, "audiences", outboundRoute.Audiences)
}
```

operator/internal/webhook/injector/pod_mutator.go
```diff
@@ -1296,41 +1311,56 @@ func (m *PodMutator) ensurePerAgentConfigMap(
...
+			}
+
+			routes = append(routes, route)
+		}
+
+		var err error
```
clawgenti: `var err error` here shadows the named return `err` declared in the function signature. The explicit `return "", "", err` on line 1356 is correct, but this pattern is fragile — if code is ever refactored to rely on named returns, this shadow will silently swallow the error. Prefer assigning to the named return directly:
```go
routesData, err = yaml.Marshal(routes)
```
(remove the `var err error` declaration above)

clawgenti (commented):
Fixes two real bugs (inline routes crash + wrong route format) and the approach is clean. A few items worth addressing before merge.

- **Silent audience truncation**: when , only  is used with no warning; operators who specify multiple audiences will silently lose the extras.
- **Shadowed `err` variable** (line 1353): `var err error` inside the `if agentRuntime != nil` block shadows the named return `err`, so a marshal failure there won't set the named return — the error is still returned via the explicit `return "", "", err`, but this is fragile and should use the named return directly.
- **`overrideRoutesConfigMapInVolumes` has no unit test**: the new `volume_builder.go` function follows the same pattern as `overrideAuthBridgeConfigMapInVolumes` but lacks a corresponding test case.

---
*Reviewed by clawgenti using the [github-pr-review skill](https://github.com/rossoctl/agent-skills/blob/main/skills/github-pr-review/SKILL.md)*

operator/internal/webhook/injector/pod_mutator.go
```diff
@@ -1296,41 +1311,65 @@ func (m *PodMutator) ensurePerAgentConfigMap(
...
+			}
+			if outboundRoute.Destination.HostRegex != "" {
+				// AuthBridge router doesn't support hostRegex - use glob pattern in host field
+				route["host"] = outboundRoute.Destination.HostRegex
+			}
+
+			// target_audience is a single string, not array
+			// Take first audience if multiple specified
+			if len(outboundRoute.Audiences) > 0 {
+				route["target_audience"] = outboundRoute.Audiences[0]
```
clawgenti, line 1353: Only the first audience is used and the rest are silently dropped — the warning log is good but callers relying on multiple audiences (possible today or via future API extensions) will get no error. Consider returning an error or at minimum documenting this limitation in the field's godoc/API type. Tracking issue #518 is referenced, but that issue should probably be linked in the AgentRuntime CRD validation too.

clawgenti (commented):
Refactors AuthBridge routes from inline config.yaml injection to a separate ConfigMap, replacing hardcoded paths with named constants — clean and well-structured change overall.

**Findings:**
- Silent truncation of multiple audiences (only `Audiences[0]` used) with a warn log but no error return — callers with multiple audiences will silently lose entries.
- `HostRegex` is mapped to the `host` field with a comment noting the semantic mismatch; this is a behavioral change with no fallback or error.

---
*Reviewed by clawgenti using the [github-pr-review skill](https://github.com/rossoctl/agent-skills/blob/main/skills/github-pr-review/SKILL.md)*

operator/internal/webhook/injector/pod_mutator.go
```diff
@@ -1296,41 +1311,65 @@ func (m *PodMutator) ensurePerAgentConfigMap(
...
+		for _, outboundRoute := range agentRuntime.Spec.Auth.Outbound {
+			route := make(map[string]interface{})
+
+			// Host or HostRegex (flat fields, not nested)
+			if outboundRoute.Destination.Host != "" {
+				route["host"] = outboundRoute.Destination.Host
+			}
+			if outboundRoute.Destination.HostRegex != "" {
+				// AuthBridge router doesn't support hostRegex - use glob pattern in host field
+				route["host"] = outboundRoute.Destination.HostRegex
```
clawgenti, line 1347: Mapping `HostRegex` directly to the `host` field is a silent semantic change: a regex pattern (e.g. `.*\.team1\.svc\.cluster\.local`) will be passed where AuthBridge expects a glob. If AuthBridge doesn't support regex syntax at all, this will silently fail to match at runtime with no observable error at admission time. Consider validating the regex-to-glob conversion or adding a more explicit warning/event.

clawgenti, line 1347: **suggestion**: `HostRegex` is being stored verbatim in the `host` field, but AuthBridge's router matches `host` as a glob, not a regex. A regex like `.*\.team1\.svc\.cluster\.local` won't match as a glob (`*.team1.svc.cluster.local` would). Consider logging a more actionable warning here (e.g. "HostRegex is not supported; treating value as a glob — ensure it uses glob syntax") or returning an error to surface the mismatch to the operator early.

clawgenti (commented):
All three items from the previous review have been addressed cleanly: shadowed `err` is gone, the audience-truncation warning is in place, and `overrideRoutesConfigMapInVolumes` has a proper unit test. One residual suggestion below.

- **`HostRegex` is silently coerced to `host`** (line 1341): a regex pattern is put in a field that AuthBridge expects as a glob. These are semantically different — a regex like `.*\.team1\.svc\.cluster\.local` will not match as a glob. The comment acknowledges the mismatch but takes no action. Worth either (a) rejecting `HostRegex` routes with a clear error/warning, or (b) documenting the expected glob syntax in the CRD field itself so callers know they must supply a glob, not a regex.

---
*Reviewed by clawgenti using the [github-pr-review skill](https://github.com/rossoctl/agent-skills/blob/main/skills/github-pr-review/SKILL.md)*

clawgenti (commented):
All previous review findings have been addressed: the `hostRegex`→glob semantic mismatch now has a clear warning log and issue #520 filed for the CRD documentation fix. CI is green across all completed checks. All checks pass. Ready for human review.

---
*Reviewed by clawgenti using the [github-pr-review skill](https://github.com/rossoctl/agent-skills/blob/main/skills/github-pr-review/SKILL.md)*

operator/internal/webhook/injector/pod_mutator.go
```diff
@@ -1296,41 +1311,71 @@ func (m *PodMutator) ensurePerAgentConfigMap(
...
+			}
+			if outboundRoute.Destination.HostRegex != "" {
+				// AuthBridge router uses glob patterns, not regex. The CRD field name
+				// "hostRegex" is misleading - it should contain glob syntax (*.example.com),
+				// not regex syntax (.*\.example\.com). Warn users about this.
+				mutatorLog.Info("hostRegex field is mapped to AuthBridge glob pattern (not regex)",
+					"namespace", namespace, "crName", crName,
+					"hostRegex", outboundRoute.Destination.HostRegex,
+					"note", "use glob syntax like '*.team1.svc.cluster.local', not regex '.*\\.team1\\.svc\\.cluster\\.local'")
+				route["host"] = outboundRoute.Destination.HostRegex
```
cwiklik, line 1347: **suggestion** — `hostRegex` is passed verbatim into AuthBridge's `host` field, which matches by **glob**, not regex. So a real regex value (e.g. `.*\.team1\.svc.cluster.local`) will silently never match at runtime — the only signal is the operator-log warning just above. The warning is a good addition; consider also validating/rejecting obvious regex metacharacters, or documenting the glob requirement on the CRD field (or renaming `hostRegex`), so users who follow the field name aren't surprised. Note the auth test asserts `route2["host"] == .*\.team1\.svc.cluster.local`, i.e. it encodes the (warned) passthrough of regex syntax into a glob field.

operator/internal/webhook/injector/pod_mutator.go
```diff
@@ -945,13 +951,18 @@ func (m *PodMutator) InjectAuthBridge(ctx context.Context, podSpec *corev1.PodSp
...
 	// original_destination_tls (strict only).
-	perAgentCMName, err := m.ensurePerAgentConfigMap(ctx, namespace, crName,
+	perAgentCMName, routesCMName, err := m.ensurePerAgentConfigMap(ctx, namespace, crName,
 		ModeEnvoySidecar, nsConfig.AuthBridgeRuntimeYAML, nsConfig, nil, mtlsMode, "", spireEnabled, agentRuntime) // bridge never runs under envoy-sidecar
 	if err != nil {
 		return false, fmt.Errorf("envoy-sidecar per-agent ConfigMap: %w", err)
 	}
 	requiredVolumes = overrideAuthBridgeConfigMapInVolumes(requiredVolumes, perAgentCMName)
 
+	// Override authproxy-routes volume if routes ConfigMap was created
```
cwiklik, line 961: **nit / question** — this envoy-sidecar path also generates `authbridge-routes-<crName>` and overrides the volume, but routes drive the proxy-sidecar token-exchange plugin, and the comment above notes the bridge "never runs under envoy-sidecar." Is the routes ConfigMap intended in this mode, or is it dead work? If it's never consumed here, gating routes generation on `ModeProxySidecar` would avoid creating an orphan ConfigMap.

cwiklik (approved):
Clean, well-tested fix for two real AuthBridge route bugs: routes now live in a separate `authbridge-routes-<crName>` ConfigMap referenced by file (fixing the "unmarshal array" crash), and use AuthBridge's flat `routing.Route` format (`host` + `target_audience`) instead of the nested CRD shape (fixing "no matching route").

Verified the cross-file consistency that matters here: the config's `routes.file` (`AuthProxyRoutesFile` = `/etc/authproxy/routes.yaml`), the volume mount (`AuthProxyMountPath` = `/etc/authproxy`), and the ConfigMap data key (`routes.yaml`) all align via the new shared constants. The override copies the volume slice (no input mutation) and sets `Optional=false`; the routes ConfigMap gets an OwnerReference for GC; and the tests were updated to fetch the separate ConfigMap and assert the flat format.

Two non-blocking notes inline (hostRegex is passed verbatim into AuthBridge's glob `host` field; the routes ConfigMap is also generated in envoy-sidecar mode where the bridge doesn't run). Multiple audiences intentionally use only the first, with a warning + tracking issue #518.

**Author**: Alan-Cha (MEMBER — maintainer) · **Areas**: Go (webhook injector) + tests · **.claude/.vscode**: none · **Commits**: 9, signed-off · **CI**: passing (16 checks).

_Assisted-By: Claude Code_

PR #535, opened 2026-09-15
Fix: Pin injected AuthBridge images to cortex v0.7.0-alpha.10

cwiklik:
## Summary

Moves the four injected AuthBridge sidecar images from cortex `v0.7.0-alpha.3` to `v0.7.0-alpha.10`, ahead of cutting operator `v0.4.0-alpha.1`.

cortex `v0.7.0-alpha.10` is tagged and all four images are published and verified pullable from `ghcr.io`.

## Pin sites

All three move together:

| File | Role |
|------|------|
| `charts/operator/values.yaml` (`defaults.images.*`) | What the chart ships |
| `operator/internal/webhook/config/defaults.go` | Compiled fallbacks that `loader.go` overlays the platform-config ConfigMap on top of, so a no-ConfigMap deploy (`make deploy`, `webhook.enable=false`) still injects a pinned tag |
| `operator/test/e2e/e2e_suite_test.go` (`sidecarImages`) | Images pulled and loaded into Kind |

## Note on the e2e list

The version-pinning guard validates only the first two files. Leaving `e2e_suite_test.go` behind would pass that guard while e2e silently exercised the older sidecars — the drift that file's own comment warns against:

> Keep tags in sync with `charts/operator/values.yaml` (`defaults.images.*`) so e2e exercises the images the chart actually ships, not `:latest`.

So this PR changes CI behavior as well as shipped metadata: e2e now loads the alpha.10 sidecars.

Also realigns `operator/test/e2e/README.md`, whose prose still described `spiffe-helper` as separately pulled and omitted `authbridge` / `authbridge-lite`.

## What this bump carries

The range crosses cortex `alpha.8`, where `proxy-init`'s egress guard broke in-cluster DNS on OpenShift — a `CLUSTER_CIDRS` default of `10.0.0.0/8` excluded OCP's 172.30/172.31 service network, so UDP/53 to the resolver hit the terminal DROP. `alpha.10` removes `CLUSTER_CIDRS` outright and derives the DNS exemption from the pod's `/etc/resolv.conf` nameservers. So this is an OpenShift fix, not a regression risk.

Caveat: **Kind e2e cannot validate that half.** Kind's service network (10.96/16) sits inside the old 10/8 default, which is why the original bug survived CI.

## Test plan

- [ ] PR-time pin guard passes — `.github/workflows/security-scans.yaml` (`Guard — AuthBridge injection images must be version-pinned`). `release.yml` carries the same two-file check but only runs on tag push, so it is not exercised here.
- [ ] Kind e2e passes against the alpha.10 sidecars
- [ ] One OCP/HyperShift run before tagging `v0.4.0-alpha.1`, to cover the DNS path Kind structurally cannot


operator/test/e2e/e2e_suite_test.go
```diff
@@ -66,10 +66,10 @@ var (
...
 	// e2e exercises the images the chart actually ships, not :latest.
 	sidecarImages = []string{
-		"ghcr.io/rossoctl/cortex/authbridge-envoy:v0.7.0-alpha.3",
-		"ghcr.io/rossoctl/cortex/authbridge:v0.7.0-alpha.3",
-		"ghcr.io/rossoctl/cortex/authbridge-lite:v0.7.0-alpha.3",
-		"ghcr.io/rossoctl/cortex/proxy-init:v0.7.0-alpha.3",
+		"ghcr.io/rossoctl/cortex/authbridge-envoy:v0.7.0-alpha.10",
+		"ghcr.io/rossoctl/cortex/authbridge:v0.7.0-alpha.10",
+		"ghcr.io/rossoctl/cortex/authbridge-lite:v0.7.0-alpha.10",
+		"ghcr.io/rossoctl/cortex/proxy-init:v0.7.0-alpha.10",
```
huang195, line 72: nit (pre-existing, and this PR is the natural place to fix it): `operator/test/e2e/README.md:18` describes this list as pulling `authbridge-envoy`, `proxy-init`, and `spiffe-helper` — which no longer matches. `spiffe-helper` isn't pulled separately (it's bundled into the combined images, as `config/defaults.go`'s own comments note), and `authbridge` and `authbridge-lite` are missing from the prose. Since the comment just above this list is what keeps it honest against `values.yaml`, worth realigning the README to the four images actually loaded.

huang195 (approved):
Verified the pin move end-to-end. Every factual claim in the description checks out; only nits below.

**What I verified independently**

- `v0.7.0-alpha.10` exists and is the newest `v0.7.0` cortex tag.
- All four images (`authbridge-envoy`, `authbridge`, `authbridge-lite`, `proxy-init`) carry that tag on ghcr, so "published and verified pullable" holds.
- **No pin site was missed.** `v0.7.0-alpha.3` appears in exactly three files on `main`, and all three are in this diff. The other two files referencing `ghcr.io/rossoctl/cortex` are not pin sites: `operator/test/e2e/README.md` is version-agnostic, and `transparent_inbound_test.go:553`'s `:latest` is a fixture input for a pre-existing init container, not an image that gets pulled.
- The `release.yml` guard reasoning is accurate: it validates `values.yaml` (loop over the four keys) and `defaults.go` (grep), and **not** `e2e_suite_test.go` — so including the e2e list here is the right call. The unanchored regex `:v[0-9]+\.[0-9]+\.[0-9]+` does match `:v0.7.0-alpha.10`.

**An upside worth stating explicitly**

This bump crosses cortex alpha.8, which is where `proxy-init`'s egress guard broke in-cluster DNS on OpenShift: the `CLUSTER_CIDRS` default of `10.0.0.0/8` excluded OCP's 172.30/172.31 service network, so UDP/53 to the resolver hit the terminal DROP. By alpha.10 that is fixed — `CLUSTER_CIDRS` was removed outright and the DNS exemption now follows the pod's `/etc/resolv.conf` nameservers, which is cluster-agnostic. The operator is already aligned: `container_builder_test.go:327` asserts enforce-redirect must *not* set `CLUSTER_CIDRS`.

So this carries an OpenShift fix rather than risking a regression. The corollary is that **Kind e2e cannot validate that half** — Kind's service net (10.96/16) sits inside the old 10/8 default, which is why the original bug survived CI. Worth one OCP/HyperShift run before cutting `v0.4.0-alpha.1`.

**Nit, not blocking:** the test-plan's first checkbox can't be satisfied by this PR — `release.yml` runs on tag push, not on PRs. The gate that actually runs here is the copy at `.github/workflows/security-scans.yaml:156`, which mirrors the same two-file scope.

Approving. I'd let the pending E2E go green before merge, since it's the only check that exercises the alpha.10 sidecars.

Issue #536, opened 2026-09-17
Fold bundle-service and token-broker into the operator image so they ship in releases


mrsabath:
## Summary

The `bundle-service` (OPA authorization policy bundles, #415) and `token-broker` (OAuth session/token management, #449) are both merged and present in the source tree at `v0.4.0-rc.1`, but neither is deployable from the release artifacts. Both ship as standalone binaries with hand-written manifests that are not templated into the Helm chart and are not built or published by CI.

Net effect for a user installing rossoctl `v0.8.0-rc.1`: they get the `authorizationpolicies` CRD and the cortex-side `opa`/`tokenbroker` plugins that consume policy bundles, but no bundle-service and no token-broker running to serve them. Testing either feature currently requires building from source.

**Resolution (agreed with @davidhadas):** fold both binaries into the existing operator image. The components were originally split out to accommodate Red Hat objections; Red Hat is no longer consuming this as their upstream, so the split now only carries cost. Keeping them opt-out is also inconsistent with the rest of the codebase, which is opt-in. Goal is that testing a feature does not require a source build.

## What was verified

Both PRs are merged to `main` and their merge commits are ancestors of the `v0.4.0-rc.1` tag:

| PR | Merge commit | In `v0.4.0-rc.1` |
|---|---|---|
| #415 Bundle Service serving OPA authorization policy bundles | `fb22549` | yes |
| #449 Add token-broker service for OAuth session and token management | `7887e96` | yes |

This is not a missing-code problem. It is a packaging and delivery gap.

### Why documentation alone cannot fix this

The images do not exist. `.github/workflows/release.yml` builds exactly two images via its matrix — `operator/Dockerfile` and `operator/cmd/agentcard-signer/Dockerfile`. Neither component is included. The standalone manifests reflect that and cannot be applied as-is:

- `operator/config/bundleservice/deployment.yaml:23` → `ghcr.io/rossoctl/bundle-service:latest`, never built or pushed
- `token-broker/deploy/02-deployment.yaml:34` → `localhost/token-broker:latest`, with the inline comment "REPLACE with the tag your build script produces"

There is also no feature gate today: `charts/operator/values.yaml` at `v0.4.0-rc.1` contains no reference to either component. The `featureGates` block covers `globalEnabled`, `envoyProxy`, `injectTools`, `perWorkloadConfigResolution`, and `skillDiscovery` only.

### Component detail

[...]

PR #490, opened 2026-07-20
Fix: regenerate token-broker go.sum for renamed cortex authlib

Ibrahim2595:
## Summary
Closing the last cross-repo `go.sum` gap from the rename, now that cortex #680 is merged.

`token-broker` pinned the **pre-rename** cortex authlib commit `v0.0.0-20260619001334-ce3417655ee8` (its go.mod declared `github.com/rossoctl/rossocortex/authbridge/authlib` — a module-path mismatch), and `go.sum` had no checksum (the placeholder was removed earlier since it couldn't be valid until cortex published the `cortex/` path).

- Bumped `github.com/rossoctl/cortex/authbridge/authlib` → post-rename `v0.0.0-20260720192358-48fb3c767572`
- Regenerated real `go.sum` via `go mod tidy` (incl. the transitive x/crypto, x/sys bumps authlib now requires)

## Verification
- `token-broker`: `go build` ✅, `go vet` ✅, `go test` ✅ (8 pkgs, 0 fail)

## Related issue(s)
Related to rossoctl/rossoctl#1972

PR #538, opened 2026-09-17
fix: repair token-broker build by bumping authlib past module rename

### LINKED ISSUE #537: token-broker module fails to build: authlib pinned to pre-rename commit with old module path
## Summary

The `token-broker` module does not build at `v0.4.0-rc.1`. Its `authlib` dependency is pinned to a commit that predates the Kagenti → Rossoctl rename, so the pinned `go.mod` still declares the old module path while `token-broker/go.mod` requires the new one. Go rejects the mismatch and the build fails before compilation.

```
$ cd token-broker && go build ./...
cmd/main.go:33:2: github.com/rossoctl/cortex/authbridge/authlib@v0.0.0-20260619001334-ce3417655ee8: parsing go.mod:
	module declares its path as: github.com/kagenti/kagenti-extensions/authbridge/authlib
	        but was required as: github.com/rossoctl/cortex/authbridge/authlib
```

## Root cause

`token-broker/go.mod` requires:

```
github.com/rossoctl/cortex/authbridge/authlib v0.0.0-20260619001334-ce3417655ee8
```

Commit `ce3417655ee8` is dated 2026-06-18, before the rename landed in cortex. At that commit `authbridge/authlib/go.mod` still declares `module github.com/kagenti/kagenti-extensions/authbridge/authlib`. The requirement path and the declared path disagree, which is a hard error in the Go module system — no amount of `go mod tidy` works around it, and `go get` also refuses to proceed while the existing requirement is unresolvable.

This went unnoticed because `token-broker` is a separate Go module and is not built by any CI workflow (`.github/workflows/release.yml` builds only `operator/Dockerfile` and `operator/cmd/agentcard-signer/Dockerfile`). Nothing in CI compiles this module, so the breakage is invisible.

## Fix

Bump the pin to a post-rename cortex commit. Verified working against the `v0.7.0-rc.1` commit `520dc8a30534303f6279f79c4af15837d1683c94`:

```
github.com/rossoctl/cortex/authbridge/authlib v0.0.0-20260915...-520dc8a30534
```

After the bump plus `go mod tidy`:

- `go build ./...` succeeds
- `go test ./...` passes across all 8 packages (`cmd`, `internal/api`, `internal/auth`, `internal/cache`, `internal/core`, `internal/oauth`, `internal/session`, `pkg/oauth`)

**There is no API drift.** `token-broker` uses exactly two symbols from `authlib` — `validation.NewLazyJWKSVerifier` and `validation.Verifier` — and `authbridge/authlib/plugins/jwtvalidation/validation/` has a zero-line diff between `ce3417655ee8` and `v0.7.0-rc.1`. Both signatures are byte-identical. The only problem is the module path, not the API.

## Follow-up: add CI coverage

The underlying reason this shipped broken is that no CI job compiles the module. Even without resolving how `token-broker` is packaged (see #536), a minimal `go build ./...` / `go test ./...` job for `token-broker/` would have caught this at the commit that introduced it. Recommend adding that regardless of the packaging outcome.

## Relationship to #536

#536 covers folding `bundle-service` and `token-broker` into the operator image so they ship in releases. This issue is narrower and independent: the module is broken on its own terms today.

It also removes the main schedule risk noted in #536. That issue flagged the stale `authlib` pin as an unknown worth timeboxing, on the assumption that three months of drift might mean real API work. It does not — the fix is a one-line `go.mod` bump. Folding `token-broker` in is therefore comparable in size to `bundle-service`, and the two need not be split across releases on that basis.

/cc @davidhadas @cwiklik

mrsabath:
## Summary

`token-broker` does not build at all on `main`. Its `authlib` dependency is pinned to `ce3417655ee8` (2026-06-18), which predates the Kagenti → Rossoctl rename. At that commit `authlib/go.mod` still declares the old module path, so Go rejects the requirement before compiling anything:

```
cmd/main.go:33:2: github.com/rossoctl/cortex/authbridge/authlib@v0.0.0-20260619001334-ce3417655ee8: parsing go.mod:
	module declares its path as: github.com/kagenti/kagenti-extensions/authbridge/authlib
	        but was required as: github.com/rossoctl/cortex/authbridge/authlib
```

## Changes

**Bump `authlib` to `2d373605612d`** (cortex `v0.7.0-alpha.10`) — chosen to match the authbridge image tags already pinned in `charts/operator/values.yaml`, rather than introducing a version skew. `go mod tidy` carries the transitive updates and raises the `go` directive to 1.26.5, which current authlib requires.

**Add a `test-token-broker` CI job.** The workflow-level `defaults.run.working-directory: operator` scopes every existing job to the operator module, so nothing in CI ever compiled `token-broker/`. That is why a total build failure shipped in `v0.4.0-rc.1` unnoticed. The new job overrides that default and runs build, vet, and tests.

## Verification

- `go build ./...` — passes
- `go vet ./...` — passes
- `go test ./...` — passes, all 8 packages (`cmd`, `internal/api`, `internal/auth`, `internal/cache`, `internal/core`, `internal/oauth`, `internal/session`, `pkg/oauth`)
- Re-verified under `-mod=readonly`, confirming the committed `go.mod`/`go.sum` are complete and CI needs no tidy step

**No API drift.** `token-broker` uses exactly two authlib symbols — `validation.NewLazyJWKSVerifier` and `validation.Verifier` — and `authbridge/authlib/plugins/jwtvalidation/validation/` is byte-identical between the old and new commits. This was purely a module-path problem.

## Notes

I checked whether the stale rename appears elsewhere: all `go.mod`/`go.sum` files across operator, rossoctl, and cortex, plus `ghcr.io/kagenti` image references. `token-broker` is the only affected module — the rename was otherwise thorough. It survived precisely because nothing built it.

Related to #536 (folding `bundle-service` and `token-broker` into the operator image). If that lands, `token-broker/go.mod` disappears in the module merge and this pin goes with it — but the CI job and the fix stand on their own until then, and the module should not be left unbuildable in the meantime.

Fixes #537


davidhadas (approved):
/lgtm

Problem: token-broker doesn't build on main. Its go.sum is missing the authlib hashes, and its go.mod pins a pre-rename cortex commit. No CI job compiles this module, so it shipped broken in v0.4.0-rc.1 and nobody noticed.

Fix: Bump the authlib pin to a post-rename commit, restore the go.sum hashes, and add a CI job that builds and tests the module.

Verified:
- The CI job overrides the workflow-level working-directory: operator, so it tests token-broker/ and not the wrong directory.
- The pin matches the authbridge images in values.yaml:247-249, so the broker and sidecar use the same library version.
- Transitive bumps are patch-level only — no k8s.io/*, no controller-runtime.

Non-blocking: go directive goes to 1.26.5 while operator/go.mod is lower — reconcile at the module merge (#536 item 2). And token-broker still has no lint job.

Approving. Closing #490, which fixed the same bug in July but pins an older commit and adds no CI.

PR #539, opened 2026-09-17
fix(deps): bump grpc to 1.83.2 to clear three CVEs

mrsabath:
## Summary

Clears all three open Dependabot alerts on `main`. All three are the same package — `google.golang.org/grpc`, an indirect dependency reached via `github.com/spiffe/go-spiffe/v2/workloadapi`:

| Severity | CVE | Fixed in |
|---|---|---|
| high | CVE-2026-84445 | 1.82.2 |
| high | CVE-2026-84304 | 1.83.1 |
| medium | CVE-2026-84303 | 1.83.1 |

Bumping grpc 1.82.1 → 1.83.2 clears all three.

## Why this instead of #533

Dependabot #533 also fixes these CVEs and its CI is fully green, but it is labeled "minor-and-patch" while actually bundling:

- `k8s.io/*` 0.36.3 → **0.37.0**
- `sigs.k8s.io/controller-runtime` 0.24.1 → **0.25.0**

Those are minor bumps of the operator's core Kubernetes dependencies. Taking them into `main` mid-RC — we are at `v0.8.0-rc.1` with GA pending — is a wider blast radius than a security fix warrants.

This PR is the CVE fix alone. The full diff is grpc plus five small transitive bumps (`cel.dev/expr`, `golang.org/x/{crypto,net,text,tools}`), with **no k8s or controller-runtime movement**. #533 remains open and can land after GA.

## Verification

Run with `CGO_ENABLED=0` and `-mod=readonly`, so the committed `go.mod`/`go.sum` are exercised as-is:

- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./...` — 19 packages pass

Four packages (`cmd`, `internal/controller`, `internal/webhook/v1alpha1`, `test/e2e`) fail in my environment for want of envtest binaries (`bin/k8s`, `/usr/local/kubebuilder/bin/etcd`) and a Kind cluster. I baselined them against unmodified `main` and they fail identically there, so the failures are pre-existing and environmental rather than caused by this bump. CI covers those suites.


davidhadas (approved):
Fix: grpc 1.82.1 → 1.83.2, which is above the fix version for all three, plus five transitive bumps that go mod tidy carries along.

